Issue #21, opened 2019-10-02
Please support Shaka Packager Raw Key


yangdu119:
Shaka Packager Raw key is used to inject custom PSSH and keys into content. This is important for entitlement licensing as the PSSHs are generated from PSSH Generator SDK.

karlg-arris:
@joeyparrish 
This would be a very important feature for those use Widevine in 'foreign key' mode

PR #63, opened 2020-10-17
Shaka Packager Raw Key support

cstranex:
See #21 - Adds support for using raw key encryption.

Retains backwards compatibility with the existing Widevine encryption by adding an `encryption_mode` directive to the encryption configuration.

Also adds drm_label and skip_encryption to inputs.

### DIFF
--- a/streamer/pipeline_configuration.py
+++ b/streamer/pipeline_configuration.py
@@ -62,6 +62,33 @@ class ProtectionScheme(enum.Enum):
   CBCS = 'cbcs'
   """AES-128-CBC mode with pattern encryption."""
 
+class ProtectionSystem(enum.Enum):
+  WIDEVINE = 'Widevine'
+  FAIRPLAY = 'FairPlay'
+  PLAYREADY = 'PlayReady'
+  MARLIN = 'Marlin'
+  COMMON = 'CommonSystem'
+
+class EncryptionMode(enum.Enum):
+  WIDEVINE = 'widevine'
+  """Widevine key server mode"""
+
+  RAW = 'raw'
+  """Raw key mode"""
+
+class RawKeyConfig(configuration.Base):
+  """An object representing a list of keys for Raw key encryption"""
+
+  label = configuration.Field(str).cast()
+  """An arbitary string or a predefined DRM label like AUDIO, SD, HD, etc.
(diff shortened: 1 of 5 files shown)


cstranex:
@googlebot I signed it!

cstranex:
@joeyparrish would you mind reviewing when you have a chance? :smile:

cstranex:
ping

cstranex:
@joeyparrish would you mind reviewing?

joeyparrish:
I'm sorry for the delay.  We have too few people on too many projects at the moment.  I'm looking at it now.

streamer/pipeline_configuration.py
```diff
@@ -62,6 +62,33 @@ class ProtectionScheme(enum.Enum):
...
+
+  RAW = 'raw'
+  """Raw key mode"""
+
+class RawKeyConfig(configuration.Base):
+  """An object representing a list of keys for Raw key encryption"""
+
+  label = configuration.Field(str).cast()
+  """An arbitary string or a predefined DRM label like AUDIO, SD, HD, etc.
+  If not specified, indicates the default key and key_id."""
```
joeyparrish, line 84: I don't understand this part.  What do you mean by default key and key id?  Are you saying the associated key and key id from this raw key entry is a default applied to all tracks without their own label?  (AUDIO, SD, HD, etc?)

joeyparrish (commented):
This is really well done.  Thank you so much for your contribution!

streamer/pipeline_configuration.py
cstranex: Yes that's correct, similar wording to the shaka-packager [documentation](https://google.github.io/shaka-packager/html/tutorials/raw_key.html#raw-key-encryption-options) but I omitted the last sentence which maybe clarifies it better?

cstranex:
> This is really well done. Thank you so much for your contribution!

Always a pleasure :)

streamer/pipeline_configuration.py
joeyparrish: Okay, now I think I understand.  How about we review docs at some point on our side for clarity, but I don't put you through any more trouble on this PR.  This seems fine for now.

Thank you again for your contribution!

Issue #37, opened 2019-10-17
Add support for ac-3 and ec-3


avelad:
Now only it's supported aac and opus. It would be good to expand the supported audio formats.

joeyparrish:
Sounds good.  Thanks for the request!

Prototype4988:
Hi, I'm very much  interested in this project. Could I get more details on it

himanshujaidka:
yes, kindly share more details about this project.

dhruvil-shah:
Hello Sir,
I explored things about working on audio formats and found ffmpeg for the same. I also tried converting a mp4 to ac3 using it and it worked perfectly fine so can I implement the same for the Shaka-Player

aryasoni98:
@avelad I would like to work on this project of GSoC 2021 .  Could I get more details on it .

avelad:
I am not a member of Shaka's team

aryasoni98:
@avelad . oh sorry, I didn't know that .

michellezhuogg:
Hello @aryasoni98 @Prototype4988 @himanshujaidka , sorry for the late reply!
We’re adding a few projects to our projects list soon, please stay tuned via our [mailing list](https://groups.google.com/g/shaka-player-gsoc). This specific project will be off the list since we got a PR for it, and we hope not to waste your time working on the same tasks. 
Please hold to start working on the projects until the GSoC program officially starts. 
If you have questions, please send it to the [Google group](https://groups.google.com/g/shaka-player-gsoc). 
Please be mindful of posting on the Github page just to claim that you are interested in working on a task, since it may spam other developers, and also confuse other students applying for GSoC at the same time. 
Please be kind and only @us for urgent issues on Github regarding your questions, since that might spam the other developers, or team members who are not GSoC mentors. Thank you!

PR #69, opened 2021-03-21
Add support for ac3 and ec3 #37

ans-negi:


ans-negi:
#37

ans-negi:
@googlebot I signed it!

joeyparrish:
This looks good to me, but I have one question: On what did you base the bitrate choices for ac3 and ec3?

ans-negi:
@joeyparrish Hey they were recommended bitrates at [FFmpeg site ](https://trac.ffmpeg.org/wiki/Encode/HighQualityAudio), so from there I took stereo values. 
And I realised for Opus and aac there was a relation between stereo and surround values, so to calculate for surround value for ac-3 and ec-3 I used same formula (160)*(6)/2*2 == 240k

avelad:
I would like more to take the values that Apple gives for Reference in HLS.
https://developer.apple.com/documentation/http_live_streaming/hls_authoring_specification_for_apple_devices

image

What do you think?

joeyparrish:
@avelad, how about this: We adjust the defaults to be more similar to Apple's recommendations, and we also create a explicit bitrate config file that exactly matches what Apple suggests, including for video.

Here's a compromise between our existing values (extrapolated from Google internal recommendations) and Apple's recommendations:

Num. Channels | AAC Bitrate | Opus Bitrate | AC3 Bitrate | EAC3 Bitrate
-- | -- | -- | -- | --
2 | 128 | 64 | 192 | 96
6 | 256 | 128 | 384 | 192

@avelad, @ans-negi, if y'all both agree, let's use the value above as the default.

avelad:
I agree!

joeyparrish (approved):

joeyparrish:
Thanks for your contribution!

PR #74, opened 2021-06-07
Add support for HEVC video codec

KarlGallagher:
Using same FFmpeg parameters as h264 for the most part but letting the encoder choose profile and level automatically.
Bitrate settings mirror vp9

tested on macOS 10.15.7 and Ubuntu 20.04 LTS (FFmpeg 4.4)

### DIFF
--- a/streamer/bitrate_configuration.py
+++ b/streamer/bitrate_configuration.py
@@ -87,6 +87,12 @@ class VideoCodec(enum.Enum):
 
   AV1 = 'av1'
   """AV1."""
+  
+  HEVC = 'hevc'
+  """HEVC, also known as h.265"""
+
+  HARDWARE_HEVC = 'hw:hevc'
+  """HEVC, with hardware encoding"""
 
   def is_hardware_accelerated(self) -> bool:
     """Returns True if this codec is hardware accelerated."""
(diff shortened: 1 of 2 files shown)


joeyparrish (changes requested):
These framerate changes seem unrelated to HEVC, and are a thing I'd prefer to discuss separately so we can merge the HEVC support more quickly.  Can you please remove them from the PR?

KarlGallagher:
Heh sorry about that :) 

Extra changes reverted

joeyparrish (changes requested):
Looks good.  Thanks!

Could you add a test case for HEVC, please?  See the "supports AV1" test case inside `codecTests()` in `tests/tests.js` for an example you can use to build a new test case.

Then, please use `run_end_to_end_tests.py` to run the tests yourself.

KarlGallagher:
@joeyparrish   I think with the way the tests are constructed right now (using shaka player to stream the output in a browser instance), any test I add will more than likely fail due to general lack of browser support  for this codec

https://caniuse.com/hevc

joeyparrish:
You're right about that.  But I believe I have a way to structure the test so that we can at least prove that encoding succeeded and that the correct codec string appears in the manifest.

joeyparrish:
I'll go ahead and merge this, then follow-up with a change to add the missing test.  I'll link to it here FYI.  Thanks!

joeyparrish (approved):

joeyparrish:
I've added a test for HEVC and a structure for future codec tests, without depending on browser support for any given codec:  https://github.com/google/shaka-streamer/commit/89254b4926894396634e085cff281e7d5f3cc2c2

Thanks again for contributing!

PR #77, opened 2021-06-15
Adding automatic frame rate pulldown

KarlGallagher:
If required output framerate is < input source frame rate adjust configuration accordingly to automatically pull down


joeyparrish (approved):

joeyparrish:
Thanks for your contribution!

Issue #47, opened 2020-01-16
Replace gsutil-based CloudNode with local authentication proxies and HTTP output support


joeyparrish:
Our [Shaka Player History live stream](https://shaka-player-demo.appspot.com/demo/#asset=https://storage.googleapis.com/shaka-live-assets/player-source.m3u8) is having an issue during playback in which we get HTTP 403 errors close to the live edge.

The cloud upload node already has logic to avoid this:

1. Snapshot manifests and playlists
2. Upload all segments, but no manifests or playlists
3. Upload snapshots of manifests and playlists

This is meant to make sure we never upload a manifest or playlist which refers to a segment that isn't fully available yet.  It doesn't seem to work well enough, though.

It's possible that there's some additional latency between uploading a segment and it being available, and that manifest/playlist uploads might need to be delayed.  But that would certainly be in conflict with future low-latency streaming work.  It also makes me wonder why the playlists aren't subject to the same latency.

I haven't carefully measured the final segments of the playlists on upload by Streamer & download by Player, so it's possible I'm making some bad assumptions about this issue.  A more detailed investigation is needed.

joeyparrish:
Our current plan is to replace CloudNode with Shaka Packager's recent HTTP upload support and a local authentication proxy.

PR #78, opened 2021-06-21
Multi period for DASH

meryacine:
## Changes:
- [`ConflictingFields`](https://github.com/meryacine/shaka-streamer/blob/5a9182737fc3176cbc76211023e2675861fc867f/streamer/configuration.py#L78) error raised when `inputs` and `multiperiod_inputs_list` fields are present at the same time in the `input_config.yaml`.
- [`MissingRequiredExclusiveFields`](https://github.com/meryacine/shaka-streamer/blob/5a9182737fc3176cbc76211023e2675861fc867f/streamer/configuration.py#L96) error raised when neither `inputs` field nor `multiperiod_inputs_list` field is present.
- [`InputConfig.__init__()`](https://github.com/meryacine/shaka-streamer/blob/5a9182737fc3176cbc76211023e2675861fc867f/streamer/input_configuration.py#L332) pre-checks for the presence or absence of these fields mentioned above and raises error accordingly.
- [`SinglePeriod`](https://github.com/meryacine/shaka-streamer/blob/5a9182737fc3176cbc76211023e2675861fc867f/streamer/input_configuration.py#L318) class added.
- [`controller._append_nodes_for_inputs_list()`](https://github.com/meryacine/shaka-streamer/blob/5a9182737fc3176cbc76211023e2675861fc867f/streamer/controller_node.py#L193) performs as what `controller.start()` was performing, `controller.start()` now calls `controller._append_nodes_for_inputs_list()` once or multiple time (based on whether the input is multiperiod or not) to append more Transcoder and Packager nodes. #change the name of this method.
- In [`ProcessStatus`](https://github.com/meryacine/shaka-streamer/blob/5a9182737fc3176cbc76211023e2675861fc867f/streamer/node_base.py#L30) `Running` is assigned to a greater value than `Finished`.
- [`periodconcat_node.py`](https://github.com/meryacine/shaka-streamer/blob/multi-period/streamer/periodconcat_node.py) added, it contains a new ThreadedNode that is appended after all the Transcoder and Packager nodes [here](https://github.com/meryacine/shaka-streamer/blob/5a9182737fc3176cbc76211023e2675861fc867f/streamer/controller_node.py#L172), it checks whether all the Packager nodes have finished or not every while (`3s`) in the [`PeriodConcatNode._thread_single_pass()`](https://github.com/meryacine/shaka-streamer/blob/5a9182737fc3176cbc76211023e2675861fc867f/streamer/periodconcat_node.py#L42), if all of them are finished, it starts the period concatenation. For DASH, `xml` module is used to parse the `.mpd` files and extract the periods and duration information, for HLS we might use `m3u8`.


meryacine:
@joeyparrish
For the last point mentioned above, when doing periods in LIVE, do we need to mention the 'start' attribute for each `Period` for dash players to play it correctly, or we can just omit it safely.

streamer/configuration.py
```diff
@@ -238,23 +247,26 @@ def get_type_name_static(type: Optional[Type],
     if type is str:
       # Call it a string, not a "str".
       return 'string'
-    elif type is list:
+    if type is list:
```
TheModMaker, line 250: I know some styles (and even Google styles) say to not use "else if" or elif when the block above returns.  But I think it is more clear to have this as an elif chain so it is clear only one of all these will be hit.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,318 @@
...
+        """If given a duration, it uses it to set the time containers."""
+        
+        self.PY: int = Y # Years
+        self.PM: int = PM  # Months
+        self.PW: int = W # Weeks
+        self.PD: int = D # Days
+        self.TH: int = H # Hours
+        self.TM: int = TM  # Minutes
+        self.TS: float = S # Seconds
+        self.dur = 'PT0S' # Duration string in iso8601
```
TheModMaker, line 97: Instead of having a separate field for this, you could just override the `__str__` method and generate it then.  That makes sure that someone editing the fields gets the correct value.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,318 @@
...
+        if self.TH:
+          self.dur += str(self.TH) + 'H'
+        if self.TM:
+          self.dur += str(self.TM) + 'M'
+        if self.TS:
+          self.dur += str(self.TS) + 'S'
+        
+        # If all the time containers were zero, write zero seconds
+        if self.dur == 'PT':
+          self.dur += '0S'
```
TheModMaker, line 129: First, you can have 0 in these fields, so you can just have every value.  Second, since we target Python3, you can use format strings for this.

```py
return f'P{self.PY}Y{self.PM}M{self.PW}W{self.PD}DT{self.TH}H{self.TM}M{self.TS}S'
```

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,318 @@
...
+       
+      def parse(self, duration: str) -> None:
+        """Parses duration string into different containers."""
```
TheModMaker, line 132: This whole thing can be done with regexes.  See:

https://github.com/google/shaka-player/blob/0146ccd69d84601fe160d608e6954bfd430360c4/lib/util/xml_utils.js#L193-L208

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,318 @@
...
+    """Concatenates multiple single-period DASH manifests into one multi-period DASH manifest."""
+    
+    class ISO8601:
+      """A helper class represeting the iso8601:2004 duration format."""
+      
+      def __init__(self, Y=0, PM=0, W=0, D=0, H=0, TM=0, S=0,
+                   duration: Optional[str] =None):
+        """If given a duration, it uses it to set the time containers."""
+        
+        self.PY: int = Y # Years
```
TheModMaker, line 90: I would suggest full names for these.  e.g. `years`, `months`.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,318 @@
...
+    self._status = ProcessStatus.FINISHED
+  
+  def _dash_concat(self) -> None:
+    """Concatenates multiple single-period DASH manifests into one multi-period DASH manifest."""
+    
+    class ISO8601:
+      """A helper class represeting the iso8601:2004 duration format."""
+      
+      def __init__(self, Y=0, PM=0, W=0, D=0, H=0, TM=0, S=0,
+                   duration: Optional[str] =None):
```
TheModMaker, line 87: Instead of having both the string and number versions, I would suggest a static method to parse from string.

```py
@staticmethod
def parse(value: str) -> ISO8601:
  ...
  return ISO8601(...)
```

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,318 @@
...
+
+import xml.etree.ElementTree as ET
+import os
+from streamer import __version__
+from streamer.node_base import NodeBase, ProcessStatus, ThreadedNodeBase
+from streamer.transcoder_node import TranscoderNode
+from streamer.packager_node import PackagerNode
+from streamer.input_configuration import Input
+from streamer.pipeline_configuration import PipelineConfig, ManifestFormat, StreamingMode
+from typing import List, Tuple, Optional
```
TheModMaker, line 25: Imports go in sections, each alphabatized.  Sections go: `__future__`, system, third-party, ours.  So it should be:

```py
import os
from typing import List, Tuple, Optional

from xml.etree import ElementTree

from streamer import __version__
...
```

streamer/periodconcat_node.py
meryacine: yes i just wanted the output to look cleaner :).
i looked up for f strings and they where supported for >= python3.6, is that ok?

streamer/configuration.py
meryacine: ok ill revert.

streamer/configuration.py
```diff
@@ -223,7 +232,7 @@ def get_subtypes(
 
     underlying = Field.get_underlying_type(type)
     if underlying is dict:
-      return cast(Tuple[Optional[Type], Optional[Type]], args)
+      return (args[0], args[1])
```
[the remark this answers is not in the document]
meryacine: Nope, i am using mypy v0.812, i will downgrade and see if it errors.

streamer/controller_node.py
```diff
@@ -196,44 +243,55 @@ def start(self, output_dir: str,
...
+    Each Transcoder and Packager nodes are paired together, the status sent to each one of them
+    in the stop(status) is the status of the other.
+    """
+    
+    for i, node in enumerate(self._nodes):
+      status: Optional[ProcessStatus] = None
+      if isinstance(node, TranscoderNode):
+        status = self._nodes[i+1].check_status()
+      elif isinstance(node, PackagerNode):
+        status = self._nodes[i-1].check_status()
```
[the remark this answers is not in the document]
meryacine: It's no more the case that there is only one transcoder and packager node, i think putting each type in a separate list will be less confusing, right?

streamer/configuration.py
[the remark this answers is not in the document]
meryacine: Tried with some random versions and them all passed.
v0.812 (what i am using)
v0.800
v0.782
v0.770
v0.750

streamer/configuration.py
```diff
@@ -458,10 +470,13 @@ def _sortable_properties(self) -> Tuple:
     """Return a tuple of sortable properties.  Implemented by subclasses."""
     raise RuntimeError('_sortable_properties missing on RuntimeMapSubclass!')
 
-  def __eq__(self, other: Any) -> bool:
+  # NOTE: ignore[override] is needed to suppress the following the mypy
+  # error: "This violates the Liskov substitution principle,
+  # See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides"
+  def __eq__(self, other: 'RuntimeMap') -> bool: # type: ignore[override]
```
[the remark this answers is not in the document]
meryacine: Well, that `Any` must have a `_sortable_properties` method for it to be comparable with `self: RuntimeMap`, so that `Any` is a `RuntimeMap`.
These magical methods are used to compare objects of type `VideoResolution` & `AudioChannelLayout` which are `RuntimeMap`s.

streamer/cloud_node.py
```diff
@@ -85,21 +85,14 @@ def check_access(bucket_url: str) -> None:
...
     if status.returncode != 0:
-      message = """Unable to write to cloud storage URL: {}
-
-Please double-check that the URL is correct, that you are signed into the
-Google Cloud SDK or Amazon AWS CLI, and that you have access to the
-destination bucket.
-
-Additional output from gsutil:
-  {}""".format(bucket_url, status.stderr)
+      message = "Unable to write to cloud storage URL: {}\n\nPlease double-check that the URL is correct, that you are signed into the\nGoogle Cloud SDK or Amazon AWS CLI, and that you have access to the\ndestination bucket.\n\nAdditional output from gsutil:\n  {}".format(bucket_url, status.stderr)
```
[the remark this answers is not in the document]
meryacine: ok.

streamer/bitrate_configuration.py
```diff
@@ -368,7 +368,8 @@ class BitrateConfig(configuration.Base):
   """
 
   video_resolutions = configuration.Field(
-      Dict[str, VideoResolution], default=DEFAULT_VIDEO_RESOLUTIONS).cast()
+      Dict[str, VideoResolution],
+      default=DEFAULT_VIDEO_RESOLUTIONS).cast()
```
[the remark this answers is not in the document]
meryacine: sorry for that.

streamer/controller_node.py
```diff
@@ -196,44 +243,55 @@ def start(self, output_dir: str,
...
-    if not self._nodes:
-      return ProcessStatus.Finished
+    
+    values: Set[int] = set([ProcessStatus.FINISHED.value])
 
-    value = max(node.check_status().value for node in self._nodes)
+    for node in self._nodes:
+      status: ProcessStatus = node.check_status()
+      if isinstance(node, (TranscoderNode, PackagerNode, PeriodConcatNode)):
+        values.add(status.value)
```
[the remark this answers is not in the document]
meryacine: `ClouldNode` and also `ExternalCommandNode`, ignoring their status because they do not tell whether the Transcodeing-Packaging pipeline is finished or not.
If all the packager and transcoder nodes are finished then we should know that the required job is done from the controller node, also `CloudNode` is never set to finished, that's another reason i don't check for it.
i think we can give the `CloudNode` a ref to all the nodes as done for the `PeriodConcatNode` to set its status to finished after the last sync.

streamer/cloud_node.py
```diff
@@ -85,21 +85,19 @@ def check_access(bucket_url: str) -> None:
...
     if status.returncode != 0:
-      message = """Unable to write to cloud storage URL: {}
-
-Please double-check that the URL is correct, that you are signed into the
-Google Cloud SDK or Amazon AWS CLI, and that you have access to the
-destination bucket.
-
-Additional output from gsutil:
-  {}""".format(bucket_url, status.stderr)
+      message = ("Unable to write to cloud storage URL: {}\n\n"
```
meryacine, line 88: @TheModMaker
Would that be ok?, i am trying to keep the indentation because it causes visual problems in vscode and i think other text editors as well.

streamer/controller_node.py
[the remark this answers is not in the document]
joeyparrish: CloudNode is on its way out: #47 

But I don't see any reason to change this method.  It seems that the old rule could still apply: `max(node.check_status().value for node in self._nodes)`.  The only difference is that now, you want the "running" status to overrule "finished".  And you've already changed their values to make that the case with `max()`.

ExternalCommandNode should not be an exception to this rule.  If it's still running, it's still generating input to the other nodes.

CloudNode still requires an exception, at least until we delete it, but that can be handled in CloudNode itself.

streamer/controller_node.py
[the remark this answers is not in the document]
joeyparrish: I don't think this change is necessary.  stop() is only called on the controller when we're stopping everything.  There are only two cases for this:

1. check_status() on controller is Finished, meaning everything is finished, so the pairs don't matter because all status are the same
2. check_status() on controller is Running or Errored, meaning something failed or we shut down early, so the pairs don't matter because we want to force-kill anything still running

streamer/input_configuration.py
```diff
@@ -315,10 +315,43 @@ def get_input_args(self) -> List[str]:
...
+    if dictionary.get('inputs') and dictionary.get('multiperiod_inputs_list'):
+      raise configuration.ConflictingFields(
+        InputConfig, 'inputs', configuration.Field(List[Input]), ['inputs', 'multiperiod_inputs_list']
+      )
+    
+    # Because these fields are not marked as required at the class level
+    # , we need to check ourselves that one of them is provided.
+    if dictionary.get('inputs') is None and dictionary.get('multiperiod_inputs_list') is None:
+      raise configuration.MissingRequiredField(
+        InputConfig, 'either inputs or multiperiod_inputs_list', configuration.Field(Dict[str, Any])
```
joeyparrish, line 353: I'm not certain about these errors where you make up configuration.Field values with bogus types like Any.  I wonder if this is why you had to add "Any" handling to the type formatter.

How about this:

1. `ConflictingFields` takes the two field names and the class and uses the names to extract the declared types
2. Create a `MissingRequiredExclusiveFields` that does the same, but with a message that says "exactly one of" ... "is required"
3. Use both field names and types in the error message strings of those error types

streamer/node_base.py
```diff
@@ -115,40 +115,40 @@ def check_status(self) -> ProcessStatus:
...
         time.sleep(1)
 
-      if self.check_status() == ProcessStatus.Running:
-        # If it's still not dead, use kill.
-        self._process.kill()
-        # Wait for the process to die and read its exit code.  There is no way
-        # to ignore a kill signal, so this will happen quickly.  If we don't do
-        # this, it can create a zombie process.
-        self._process.wait()
+        if self.check_status() == ProcessStatus.RUNNING:
```
joeyparrish, line 135: No need to indent this further.  It was equally correct before.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+    # to one TranscoderNode and its corresponding PackagerNode in each of these lists.
+    self._transcoder_nodes: List[TranscoderNode] = []
+    self._packager_nodes: List[PackagerNode] = []
+    
+    # We don't need to save CloudNode(s) or ExternalCommandNode(s).
+    # We will start the concatenation once all the PackagerNode(s) are finished.
+    for node in all_nodes:
+      if isinstance(node, TranscoderNode):
+        self._transcoder_nodes.append(node)
+        self._period_inputs_list.append(node._inputs)
```
joeyparrish, line 54: It's poor form to access a private member of another class.  Also, I don't see this being used anywhere.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+    for i, packager_node in enumerate(self._packager_nodes):
+      status: ProcessStatus = packager_node.check_status()
+      if status == ProcessStatus.RUNNING:
+        return
+      elif status == ProcessStatus.ERRORED:
+        raise RuntimeError(
+          "PeriodConcatNode: 'PackagerNode#{} Errored, Concatenation is stopped.'".format(i)
+        )
+    
+    if(ManifestFormat.DASH in self._pipeline_config.manifest_format):
```
joeyparrish, line 73: style nit: Don't use parens unless you need to span two lines

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+    if(ManifestFormat.HLS in self._pipeline_config.manifest_format):
+      self._hls_concat()
+    
+    self._status = ProcessStatus.FINISHED
+  
+  def _dash_concat(self) -> None:
+    """Concatenates multiple single-period DASH manifests into one multi-period DASH manifest."""
+    
+    class ISO8601:
+      """A helper class represeting the iso8601:2004 duration format."""
```
joeyparrish, line 85: typo: representing

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+      """A better interface for the Element.findall() method"""
+      full_path = ''
+      for tag in args:
+        full_path+='{'+dash_namespace+'}'+tag+'/'
+      return elem.findall(full_path[:-1])
+    
+    # We will add the time of each period on this container to set that time later
+    # as 'mediaPresentationDuration' for the MPD tag for VOD.
+    # For LIVE we will use this object to determine the start time of the next period.
+    total_time: ISO8601 = ISO8601()
```
joeyparrish, line 159: You don't need to do this.  If you move the mediaPresentationDuration of each mpd into the Period element's duration attribute, you can omit mediaPresentationDuration from the combined mpd element.  Players must be able to handle this, so you can skip all the duration parsing above.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Concatenates inputs into periods by creating a master DASH/HLS file."""
+
+import os
+import re
+from typing import List, Tuple
+from xml.etree import ElementTree
```
joeyparrish, line 20: Please add `xml.etree` to the list of dependencies in `docs/source/prerequisites.rst`, as well as the `install_requires` list in `setup.py`.

joeyparrish (changes requested):

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+    ElementTree.register_namespace('', dash_namespace)
+    
+    
+    # Write the period concat to the outpud_dir.
+    with open(os.path.join(
+        self._output_dir,
+        self._pipeline_config.dash_output), 'w') as master_dash:
+      contents = "<?xml version='1.0' encoding='UTF-8'?>\n"
+      contents += "<!--Generated with https://github.com/google/shaka-packager -->\n"
+      contents += "<!--Made Multi-Period with https://github.com/google/shaka-streamer version {} -->\n".format(__version__)
```
joeyparrish, line 245: I love this so much, I wish it were there on single-period streams, too.

Since we're moving toward having Packager write directly to cloud storage, maybe we can add a feature to Packager where we can specify an additional comment to include.  Then we can ask Packager to add Streamer branding directly to all output files.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+
+class PeriodConcatNode(ThreadedNodeBase):
+  """A node that concatenates multiple DASH manifests and/or HLS playlists
+  when the input is a multiperiod_inputs_list.
+  """
+  
+  def __init__(self,
+               pipeline_config: PipelineConfig,
+               all_nodes: List[NodeBase],
+               outpud_dir: str) -> None:
```
joeyparrish, line 37: Typo: output_dir

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+      self._packager_nodes[0]._output_dir,
+      self._pipeline_config.dash_output)).getroot()
+    
+    # Remove the Period element in that MPD.
+    ref_mpd.remove(find(ref_mpd, 'Period'))
+    
+    for i, packager_node in enumerate(self._packager_nodes):
+      
+      dash_file_path = os.path.join(
+        packager_node._output_dir,
```
joeyparrish, line 175: Make output_dir public on PackagerNode (remove the leading underscore)

streamer/controller_node.py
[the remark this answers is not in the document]
joeyparrish: Here's a suggestion:

1. Pass all packager nodes to both PeriodConcatNode and CloudNode constructors.
2. Add a static method to NodeBase to query the combined status of a list of nodes, based on the `max()` comprehension.
3. Make ControllerNode use the new static method here to compute status of all nodes.
4. PeriodConcatNode can use the new static method to wait for all PackagerNodes to have a Finished status, then combine manifests/playlists, then shut down.
5. If CloudNode sees Finished across PackagerNodes, it can make one additional upload pass, then shut down.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+class PeriodConcatNode(ThreadedNodeBase):
+  """A node that concatenates multiple DASH manifests and/or HLS playlists
+  when the input is a multiperiod_inputs_list.
+  """
+  
+  def __init__(self,
+               pipeline_config: PipelineConfig,
+               all_nodes: List[NodeBase],
```
joeyparrish, line 36: Instead of all_nodes, could this just be packager_nodes?

joeyparrish (changes requested):

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+        period.attrib['start'] = str(total_time)
+        #total_time += ISO8601.parse(period.attrib['where to get the period time from???'])
+        
+      # For multi segment media
+      for segment_template in findall(
+        period,
+        'AdaptationSet',
+        'Representation',
+        'SegmentTemplate'):
+        path_to_segement = os.path.relpath(packager_node._output_dir, self._output_dir)
```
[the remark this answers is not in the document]
meryacine: that's way neater, thanks.

streamer/periodconcat_node.py
```diff
@@ -0,0 +1,256 @@
...
+    
+    dash_namespace = 'urn:mpeg:dash:schema:mpd:2011'
+    
+    def find(elem: ElementTree.Element, *args: str) -> ElementTree.Element:
+      """A better interface for the Element.find() method.
+      Use it only if it is guaranteed that the element we are searching for is inside,
+      Otherwise it will raise an AssertionError."""
+      full_path = ''
+      for tag in args:
+        full_path+='{'+dash_namespace+'}'+tag+'/'
```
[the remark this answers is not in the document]
meryacine: okay , I rather be reading it from the sample dash file generated by packager.

streamer/periodconcat_node.py
meryacine: `xml` is from the standard library.
https://docs.python.org/3/library/xml.etree.elementtree.html
we don't need to list it as a dependency right?

streamer/periodconcat_node.py
meryacine: I thought i might need it for the HLS, but i will remove it for now.

streamer/periodconcat_node.py
joeyparrish: I stand corrected!  No need to call it a dependency.

streamer/input_configuration.py
```diff
@@ -315,10 +315,41 @@ def get_input_args(self) -> List[str]:
...
+    
+    We need these checks before passing the input dictionary to the configuration.Base constructor,
+    because it does not check for this 'exclusive or-ing' relationship between fields.
+    """
+
+    assert isinstance(dictionary, dict), """Malformed Input Config File,
+    See some examples at https://github.com/google/shaka-streamer/tree/master/config_files.
+    """
+
+    if dictionary.get('inputs') and dictionary.get('multiperiod_inputs_list'):
```
joeyparrish, line 344: This uses "truthiness", so empty lists will not be matched.  This means this condition would allow:

```yaml
inputs: []
multiperiod_inputs_list:
 - inputs:
   - name: Sintel.2010.4k.mkv
     media_type: video
```

I would prefer to treat this as conflicting fields, so let's use "is not None" here.

streamer/input_configuration.py
```diff
@@ -315,10 +315,41 @@ def get_input_args(self) -> List[str]:
...
+      raise configuration.ConflictingFields(
+        InputConfig, 'inputs', 'multiperiod_inputs_list')
+    
+    # Because these fields are not marked as required at the class level
+    # , we need to check ourselves that one of them is provided.
+    if dictionary.get('inputs') is None and dictionary.get('multiperiod_inputs_list') is None:
```
joeyparrish, line 350: This uses "is None", so empty lists will not be matched.  This means this condition would allow:

```yaml
inputs: []
```

I would prefer to treat this as missing fields, so let's drop use "is None" here and just use truthiness.

Combined with the above suggestion, we would require exactly one field, and that the exactly one field must be non-empty.

We should add tests for these various cases, too.

joeyparrish (changes requested):
Looks great otherwise.  Thanks!

streamer/input_configuration.py
meryacine: Sure, that would be better.

streamer/input_configuration.py
meryacine: I did want such malformed field error or other types of errors to be caught in the `Base` class constructor instead, that would provide more descriptive error messages about the cause of the error.
But just found out that empty lists between all other types indeed pass this error checking.

Can we add some check in `Base._check_and_convert_type`  that if a list is passed by the user, it must not be empty ??

meryacine: I tried adding this error for the length of the user-inputted list and ran the tests, some `resolution` fields were given as an empty list, after putting some dummy resolutions inside them, the tests passed.

What do you think about this approach?

tests/tests.js
```diff
@@ -1178,3 +1212,38 @@ function muxedTextTests(manifestUrl, format) {
...
+    const pipelineConfigDict = {
+      'streaming_mode': 'vod',
+      'resolutions': ['144p'],
+      'audio_codecs': ['aac'],
+      'video_codecs': ['h264'],
+    };
+
+    await startStreamer(inputConfigDict, pipelineConfigDict);
+    await player.load(manifestUrl);
+    const duration = player.seekRange().end
```
meryacine, line 1243: @joeyparrish should i be using this to determine the time??
also is there any other way to check for this, like some function that tells the number of periods in the manifest?

streamer/input_configuration.py
joeyparrish: I think you're right.  We should check for empty lists generally.  Possibly empty strings, too.  But let's put that in a separate PR.

I'm thinking of a property on Field that indicates whether empty is acceptable, defaulting to False.

tests/tests.js
```diff
@@ -316,6 +318,38 @@ function errorTests() {
...
+
+    await expectAsync(startStreamer(inputConfig, pipeline_config))
+        .toBeRejectedWith(jasmine.objectContaining({
+          error_type: 'ConflictingFields',
+          field_name: 'inputs',
+        }));
+  });
+
+  it('fails when neither "inputs" nor "multiperiod_inputs_list" is given', async() =>{
+    const inputConfig = {}
```
joeyparrish, line 341: style nit: add a semicolon at the end of this statement

tests/tests.js
```diff
@@ -316,6 +318,38 @@ function errorTests() {
           field_name: 'content_id',
         }));
   });
+
+  it('fails when both "inputs" and "multiperiod_inputs_list" are given', async() =>{
+    const inputConfig = getBasicInputConfig();
+    inputConfig.multiperiod_inputs_list = [
+      getBasicInputConfig(),
+      getBasicInputConfig(),
+    ]
```
joeyparrish, line 327: style nit: add a semicolon at the end of this statement

tests/tests.js
```diff
@@ -316,6 +318,38 @@ function errorTests() {
           field_name: 'content_id',
         }));
   });
+
+  it('fails when both "inputs" and "multiperiod_inputs_list" are given', async() =>{
```
joeyparrish, line 322: style nit: add a space between `=>` and `{`

tests/tests.js
```diff
@@ -316,6 +318,38 @@ function errorTests() {
...
+    };
+
+    await expectAsync(startStreamer(inputConfig, pipeline_config))
+        .toBeRejectedWith(jasmine.objectContaining({
+          error_type: 'ConflictingFields',
+          field_name: 'inputs',
+        }));
+  });
+
+  it('fails when neither "inputs" nor "multiperiod_inputs_list" is given', async() =>{
```
joeyparrish, line 340: style nit: add a space between `=>` and `{`

tests/tests.js
joeyparrish: There is no way to check the number of periods in Shaka v3.  In Shaka v2, we kept periods separate, but now we combine them in the player when we parse DASH.

joeyparrish (changes requested):
Looks pretty good!  Just a few nits, mostly style.

joeyparrish (approved):
Well done!  Congrats!

PR #81, opened 2021-07-01
Slight speed up of vp9 encoding

KarlGallagher:
Faster vp9 encoding, whilst still maintaining balance between speed/quality.

In my testing, for a given file, this results in ~20% reduction in encoding time.
The subjective difference in output quality appears negligible.


joeyparrish (approved):
Nice!  Thank you!

PR #83, opened 2021-07-20
Multi period for HLS

meryacine:
This PR adds the support for multi-period concatenation for HLS, using `#EXT-X-DISCONTINUITY` between periods.

The new file `m3u8_concater.py` has the classes used to parse the m3u8 files and reduce multiple files of them into one concatenated file.

`HLSConcater` is just an API that is exposed to the `PeriodConcatNode` for abstraction.

`MasterPlaylist` is a class representing a master hls playlist, the parsing part happens in the `__init__` method, creates a `MediaPlaylist` object for each media(video,text,audio) playlist found in this master playlist, the stream attributes for each playlist is passed to the `MediaPlaylist` object and stored in it.

`MediaPlaylist` is a class representing an hls media playlist, the playlist gets parsed in the `__init__` method, the biggest part of the playlist is saved mostly unchanged in the `self.content` variable, but the media segment paths get updated to point to the new relative path of the media segment.

The playlist concatenation methods are the static methods in the `MediaPlaylist` class (`concat_sub`, `concat_aud`, `concat_vid`, `concat_aud_only`(to handle the audio-only stream cases)).

No period concatenation is performed if an inconsistent media in periods is detected(e.g. one period has video and one does not).

Things that need addressing:
1- When Shaka Player gets an HLS master playlist containing unsupported and supported audio codecs, it breaks, this happens with:
- audio-only playlists with and without concatenation,
- video+audio playlists that have been concatenated, but does't happen for the un-concatenated playlists with the same set of codecs, this may have to do with the discontinuities inserted between periods.

2- When Shaka Packager receives an audio-only period to package, it normally creates a stream variant(stream-inf) for each media(audio) playlist generated, but when this audio-only input has one subtitles playlist along with the audio, Shaka-Packager doesn't create the stream-inf playlists, for the current implementation, this will create an error in the HLS concater, we can escape this period as a solution if it has no stream-infs, periods with no stream-inf we can't know its bitrates and codecs and isn't playable in Shaka Player.


streamer/periodconcat_node.py
```diff
@@ -38,6 +41,39 @@ def __init__(self,
...
+        print("Period#{} has {}video and has {}audio while Period#1 "
+              "has {}video and has {}audio.".format(i + 1, 
+                                                    "" if has_vid else "no ",
+                                                    "" if has_aud else "no ",
+                                                    "" if fp_has_vid else "no ",
+                                                    "" if fp_has_aud else "no "))
+        print("\nHINT:\n\tBe sure that either all the periods have video or all do not,\n"
+              "\tand all the periods have audio or all do not, i.e. don't mix videoless\n"
+              "\tperiods with other periods that have video that is for the concatenation\n"
+              "\tto be performed successfully.\n")
```
joeyparrish, line 74: English nit:  This sentence is too long.  Please end this sentence after "other periods that have video."  If you want to say that this is necessary for concatenation, make that a second sentence.

streamer/periodconcat_node.py
```diff
@@ -38,6 +41,39 @@ def __init__(self,
...
+      has_vid, has_aud = False, False
+      for output_stream in packager_node.output_streams:
+        if isinstance(output_stream, VideoOutputStream):
+          has_vid = True
+        elif isinstance(output_stream, AudioOutputStream):
+          has_aud = True
+      if has_vid != fp_has_vid or has_aud != fp_has_aud:
+        # Overwrite the start and the stop methods.
+        setattr(self, 'start', lambda: None)
+        setattr(self, 'stop', lambda _=None: None)
```
joeyparrish, line 63: I don't like overwriting these methods.  How about you set a flag like self._fatal_error = True and then throw a RuntimeError at the top of _thread_single_pass if it's set?

streamer/periodconcat_node.py
```diff
@@ -130,9 +168,25 @@ def find(elem: ElementTree.Element, *args: str) -> ElementTree.Element:
...
       contents += ElementTree.tostring(element=concat_mpd, encoding='unicode')
       master_dash.write(contents)
-      
+  
   def _hls_concat(self) -> None:
-    """Concatenates multiple HLS playlists with #EXT-X-DISCONTINUITY."""
+    """Concatenates multiple HLS playlists using #EXT-X-DISCONTINUITY."""
+    
+    # Initialize the HLS concater with a sample Master HLS playlist and
+    # the output direcotry of the concatenated playlists.
```
joeyparrish, line 176: typo: directory

streamer/periodconcat_node.py
```diff
@@ -130,9 +168,25 @@ def find(elem: ElementTree.Element, *args: str) -> ElementTree.Element:
...
+    hls_concater = HLSConcater(os.path.join(self._packager_nodes[0].output_dir,
```
joeyparrish, line 177: It took me a minute to read and understand this join.  Can you make it easier to read by assigning this path to a local variable with an explanatory name?  Perhaps something like first_playlist_path?

streamer/periodconcat_node.py
```diff
@@ -130,9 +168,25 @@ def find(elem: ElementTree.Element, *args: str) -> ElementTree.Element:
...
+                                            self._pipeline_config.hls_output),
+                               self._output_dir)
+    
+    for packager_node in self._packager_nodes:
+      hls_concater.add(os.path.join(packager_node.output_dir,
```
joeyparrish, line 182: Same here.  Please assign the path to a local var for clarity.

joeyparrish (changes requested):
I still need to read m3u8_concater.py, but I haven't been able to make time yet.  Wanted to go ahead with feedback on the other files.

streamer/periodconcat_node.py
meryacine: ah, sure, 
Can we not error when we see the flag?, i think of supporting multiple inputs even if it's not going to be concatenated.
Can we just set the status to finished in the _thread_single_pass.

joeyparrish: We don't intend to support it in the player, so we don't intend to support that scenario here, either.  I think it would be best to guide people toward more reasonable, widely-supported structures in media, until there's a compelling use-case for us to broaden support.  We haven't heard one yet.

meryacine: What i meant is that the user can use one streamer instance to process multiple inputs, instead of instantiating multiple instances of streamer, for LIVE for example, since there is no concatenation support, i just didn't append the concater node [here](https://github.com/google/shaka-streamer/blob/master/streamer/controller_node.py#L170), so the user can push multiple live streams from the same streamer instance.

meryacine: oh i think i miss understood you, ok let's raise an error till someone requests not to.

joeyparrish (changes requested):
Still haven't finished reading m3u8_concater.  It is quite long and will take me several passes, I think.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,840 @@
...
+  line += ','
+  # Search for all KEY=VALUE,
+  matches: List[str] = re.findall(
+    r'[-A-Z]+=(?:[-A-Z]+|[\d\.]+|\d+x\d+|\"[^\"]*\"),',
+    line)
+  # NOTE: This regex r'[-A-Z0-9]+=(?:[-A-Z0-9]+|\d*\.?\d*|\d+x[0-9A-F]+|\"[^\"]*\"),'
+  # better follows the attribute rules in
+  # https://datatracker.ietf.org/doc/html/draft-pantos-http-live-streaming-23#section-4.2
+  for match in matches:
+    key, value = match[:-1].split('=', 1)
```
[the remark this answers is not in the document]
meryacine: I will need this stream_info(attributes) to construct the new file, some HLS attributes are quoted(e.g. CHANNLES) and some are not(e.g. TYPE), i can't know latter when writing the file which to quote and which not to.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,840 @@
...
+def _search_attributes(line: str) -> Dict[str, str]:
+  """Extracts attributes from an m3u8 #EXT-X tag to a python dictionary."""
+  
+  attributes: Dict[str, str] = {}
+  line = line.strip().split(':', 1)[1]
+  # For a tighter search, append ',' and search for it in the regex.
+  line += ','
+  # Search for all KEY=VALUE,
+  matches: List[str] = re.findall(
+    r'[-A-Z]+=(?:[-A-Z]+|[\d\.]+|\d+x\d+|\"[^\"]*\"),',
```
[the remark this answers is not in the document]
meryacine: Clever, thanks :)

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,840 @@
...
+  
+  return '"' + string + '"'
+
+def _unquote(string: str) -> str:
+  """Removes the double quotes that surround a string."""
+  
+  return string[1:-1]
+
+def _codec_resolution_channels_regex(filename: str) -> Tuple[str, str]:
+  """Gets the codec name and (resolution name or channel count) from a file name,
```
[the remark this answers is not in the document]
meryacine: An audio playlist as a media tag will have no information about its codec, we will have to get that from a stream variant, and we still won't know which codec of the(say 5 audio codecs) this playlist represents, i think shaka player knows such information by playing the media segments and check its codecs, i don't know if i can do such thing here, but this was the simplest way came to my mind, whose survival is bound to the steamer's output format :(.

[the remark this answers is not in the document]
joeyparrish: No, we don't get to know that in the player, either.  HLS sucks for not providing that kind of metadata.

How about getting it from the Packager nodes somehow?  What we really need is a map of playlist filenames to the output stream's characteristics.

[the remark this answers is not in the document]
meryacine: I find that way reasonable. alright.

streamer/packager_node.py
```diff
@@ -112,7 +112,7 @@ def start(self) -> None:
       # Log by writing all Packager output to a file.  Unlike the logging
       # system in ffmpeg, this will stop any Packager output from getting to
       # the screen.
-      stdout = open('PackagerNode.log', 'w')
+      stdout = open('PackagerNode.log', 'a')
```
joeyparrish, line 115: The log files could have a number added, based on the period, such as `PackagerNode-1.log`.  That would be a nice solution to the problem of overwriting them, I think.

streamer/transcoder_node.py
```diff
@@ -142,13 +142,13 @@ def start(self) -> None:
 
         args += [output_stream.ipc_pipe.write_end()]
 
-    env = {}
+    stderr = None
```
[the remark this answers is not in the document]
joeyparrish: See my comment on the PackagerNode logs.  Let's go back to using `FFREPORT`, so we get separate log verbosity for stderr (just progress) and log file (more verbose).  But let's also change the file name to indicate the period.

joeyparrish (changes requested):

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+    and store these headers in their classes respectively.
+    
+    We use the `output_location` to re-evaluate a media segment paths, and also write
+    the concatenation result to this `output_location`.
+    """
+    
+    # Save common master playlist header, this will call
+    # MediaPlaylist.extract_header() to save the common
+    # media playlist header as well.
+    MasterPlaylist.extract_headers(sample_master_playlist_path)
```
joeyparrish, line 895: It bothers me that these are stored statically on the class.  It means that an application using the Python API can't create multiple independent controllers doing concatenation if the headers would need to be different.

What if, instead, the headers (both master & media headers) are stored on the HLSConcater instance, and passed to .write()?

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+      # We will get the language from the stream_info because the stream information
+      # is provided by Packager.  We might have mixed 3-letter and 2-letter format
+      # from the Streamer, but Packager reduces them all to 2-letter language tag.
+      self.language = _unquote(self.stream_info.get('LANGUAGE', '"und"'))
+    elif isinstance(output_stream, TextOutputStream):
+      self.language = _unquote(self.stream_info.get('LANGUAGE', '"und"'))
+  
+  def write(self, dir_name: str) -> None:
+    file_path = os.path.join(dir_name, _unquote(self.stream_info['URI']))
+    with open(file_path, 'w') as media_playlist:
```
joeyparrish, line 175: There are other places where "media_playlist" is a MediaPlaylist instance.  Here, it's a file instance.  That makes it confusing when searching the code for media_playlist.write(), because they mean two different things in these two contexts.  Could you please rename this file instance to something else?

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+    """Adds a master playlist to the HLSConcater object, to be concatented in
+    order when HLSConcater.concat() is called.
+    """
+    
+    self._all_master_playlists.append(
+        MasterPlaylist(master_playlist_path,
+                       self._output_location,
+                       packager_node))
+  
+  def concat(self) -> None:
```
joeyparrish, line 911: I'm not sure why concat() and write() are separate steps in this class, or why the resulting master playlist is saved to a member variable.  These should be combined, I think, and the master playlist should be a local variable.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+        stream_type = media_playlist.stream_info.get('TYPE', 'STREAM-INF')
+        if stream_type == 'SUBTITLES':
+          txt_playlists.append(media_playlist)
+        elif stream_type == 'AUDIO':
+          aud_playlists.append(media_playlist)
+        elif stream_type == 'STREAM-INF':
+          var_playlists.append(media_playlist)
+        else:
+          # TODO: We need a case for CLOSED-CAPTIONS(CC).
+          raise RuntimeError("TYPE={} is not regonized.".format(stream_type))
```
joeyparrish, line 849: typo: recognized

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+      all_txt_playlists.append(txt_playlists)
+      all_aud_playlists.append(aud_playlists)
+      all_var_playlists.append(var_playlists)
+      durations.append(master_playlist.duration)
+    
+    master_hls = MasterPlaylist()
+    master_hls.playlists.extend(
+        MediaPlaylist.concat_sub(all_txt_playlists, durations))
+    
+    if all(not MediaPlaylist.inf_is_vid(inf_pl) for inf_pl in all_var_playlists):
```
joeyparrish, line 860: I'm having trouble reading this.  What's "inf" short for?  Variable names make a big difference in readability and maintainability.

Since "inf_pl" is an element from the list "all_var_playlists", could this fairly be called "var_playlist" instead?

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+      # If the user has an audio-only content, the concatenation will go a little
+      # bit different to produce the desired output.
+      master_hls.playlists.extend(
+          MediaPlaylist.concat_aud_only(
+            all_aud_playlists,
+            all_var_playlists,
+            durations))
+    else:
+      master_hls.playlists.extend(
+          MediaPlaylist.concat_aud(all_aud_playlists))
```
joeyparrish, line 872: Why don't you need the duration for audio?

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+    # We will fill this map, file names to (OutputStream)s and will give it out
+    # to each media playlist we initialize.  Each media playlist then will pick
+    # the right output stream for itself.
+    streams_map: Dict[str, OutputStream] = {}
+    
+    # Add a mapping between single segment file names and their output stream.
+    streams_map.update({
+        outstream
+        .get_single_seg_file()
+        .write_end(): outstream
```
joeyparrish, line 726: Will these fit on one line?  I think I would find it easier to read that way.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+        for outstream in packager.output_streams
+      })
+    # Add another mapping between the first segment in multi-segment file names
+    # and their corresponding output streams.
+    streams_map.update({
+        outstream
+        .get_media_seg_file()
+        .write_end()
+        .replace('$Number$', '1'): outstream
```
joeyparrish, line 735: Same here.  Does this fit on one line?

joeyparrish (changes requested):
Just 700 lines left to review... :-)

streamer/m3u8_concater.py
meryacine: For text streams we use the durations to substitute for missing text streamer(in the worst case) in some periods, but ideally we don't need the durations anywhere else other than the with stream variants, because the durations are used to calculated the average bitrate of the concatenated playlist.

streamer/m3u8_concater.py
meryacine: Oh, sorry about that, i just renamed all the `inf`s to `var`s today but overlooked this one.
I was calling a playlist that is a stream variant(`#EXT-X-STREAM-INF`) an `inf` playlist.

streamer/m3u8_concater.py
meryacine: This does.

streamer/m3u8_concater.py
meryacine: But this does not, can we make an exception for this :), ends at col.87

streamer/m3u8_concater.py
meryacine: Yeah, we can do this.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+      concat_txt_playlist.target_duration = MediaPlaylist._max_targer_dur(
+          non_nones(optional_txt_playlists))
+      for i, optional_txt_playlist in enumerate(optional_txt_playlists):
+        if optional_txt_playlist:
+          # If a playlist is there, append it.
+          concat_txt_playlist.content += optional_txt_playlist.content
+        else:
+          # If the no playlist were found for this period,
+          # Create a time gap filled with dummy data for the period's duration.
+          dummy = ',\ndata:text/plain,NO {} SUBTITLES\n'.format(lang.upper())
```
meryacine, line 414: I haven't actually tested whether this renders on the screen normally or not, i have a problem with subtitles in HLS, their streams are being dropped in chrome and firefox too. Will try to fix it and report back.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+      for codec in codecs:
+        for lang in langs:
+          # If this language for this codec in this period has no media playlists
+          # with a for any channel layout, this means that the language itself
+          # is missing.  We will try to find a substitution for it.
+          if not len(codec_lang_division[codec][lang]):
+            aud_playlist_options = [codec_lang_division[codec][lang][0] for
+                                    lang in langs
+                                    if len(codec_lang_division[codec][lang])]
+            sub_lang = MediaPlaylist._fit_missing_lang(aud_playlist_options,
```
meryacine, line 493: This substitution is not as big as the one you do in the player(based on channel count, bitrate, sample rate, etc..).
It is only based on the language only for simplicity, so if some language has no playlists at all with any channel layout, we query `_fit_missing_lang()` with the all the media playlists we have for that period and a language to fit upon, the method will return a best fit based on the language only.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+        if optional_txt_playlist:
+          # If a playlist is there, append it.
+          concat_txt_playlist.content += optional_txt_playlist.content
+        else:
+          # If the no playlist were found for this period,
+          # Create a time gap filled with dummy data for the period's duration.
+          dummy = ',\ndata:text/plain,NO {} SUBTITLES\n'.format(lang.upper())
+          # Break the period's duration into target duration count and remains,
+          # because the target duration might be a prime number, this might
+          # cause an accumulated rounding error.
```
joeyparrish, line 417: This is fine, BTW.

joeyparrish (changes requested):
Still reviewing concatenation of audio and video, but I've read and commented on the rest now.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,955 @@
...
+      division[codec] = {}
+      for resolution in resolutions:
+        division[codec][resolution] = []
+    
+    # In each period do the following:
+    for vid_playlists in all_vid_playlists:
+      # Initialize a mapping between video codecs and a list of resolutions available.
+      codec_division: Dict[VideoCodec, List[MediaPlaylist]] = {}
+      for codec in codecs:
+        codec_division[codec] = []
```
joeyparrish, line 646: No need to change this in this PR, but just as a tip, you might consider if [collections.defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) might be a useful tool for this pattern.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,955 @@
...
+      # For every video playlist in this period, append it to the matching codec.
+      for vid_playlist in vid_playlists:
+        assert isinstance(vid_playlist.codec, VideoCodec)
+        codec_division[vid_playlist.codec].append(vid_playlist)
+      for codec in codecs:
+        # Sort the variants from low resolution to high resolution.
+        codec_division[codec].sort(key=lambda pl: pl.resolution)
+        for i, resolution in enumerate(sorted(resolutions)):
+          division[codec][resolution].append(
+              # Append the ith resolution if found, else, append
```
joeyparrish, line 656: So you're not comparing the actual resolutions?  I'm thinking about this abstractly here at first, with something like:

[360p, 720p, 1080p]  +  [720p, 1080p, 4k]

Going by index, you'd end up with something misaligned like:

[360p + 720p, 720p + 1080p, 1080p + 4k]

Instead of something like:

[360p + 720p, 720p + 720p, 1080p + 1080p, 1080p + 4k]

At least in Player, we have to contend with these scenarios.  But I guess here, we don't, since the bottom rungs of the output from Streamer will always align since they came from the same pipeline config, and we only cut off the top rungs based on the input res.  Is that accurate?

Could you explain this in the comment a little?  (If I understood correctly, add to the comment that the lower resolutions will always align because the outputs shared a pipeline config, so this indexing is always safe.)

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,955 @@
...
+              [key + '=' + value for
+               key, value in media_playlist.stream_info.items()]) + '\n'
+      content += '\n'
+      # Then write #EXT-X-STREAM-INF media playlists.
+      for media_playlist in self.playlists:
+        if media_playlist.stream_info.get('TYPE') is None:
+          media_playlist.write(dir_name)
+          # We don't write the URI in the attributes of a stream
+          # variant playlist.  Pop out the URI.
+          uri = _unquote(media_playlist.stream_info.pop('URI'))
```
joeyparrish, line 781: This is really cool.  I had no idea you could "pop" from a dictionary in Python.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,955 @@
...
+      # codec/language.
+      for aud_playlist in aud_playlists:
+        assert isinstance(aud_playlist.codec, AudioCodec)
+        codec_lang_division[aud_playlist.codec][aud_playlist.language].append(
+            aud_playlist)
+      # Sort and replace the missing languages in the codec_lang_division map.
+      for codec in codecs:
+        for lang in langs:
+          # If this language for this codec in this period has no media playlists
+          # with a for any channel layout, this means that the language itself
```
joeyparrish, line 487: typo: maybe remove "with a"?  Guessing at your original intent.

streamer/m3u8_concater.py
joeyparrish: Sounds good.  You should be able to count on similar channels for the same reason as similar resolutions in video: everything went through the same pipeline config.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,955 @@
...
+    
+    return concat_aud_playlists
+  
+  @staticmethod
+  def concat_aud_only(all_aud_playlists: List[List['MediaPlaylist']],
+                      all_var_playlists: List[List['MediaPlaylist']],
+                      durations: List[float]) -> List['MediaPlaylist']:
+    """Concatenates audio only periods with other audio only periods."""
+    
+    # Pair audio streams with their equivalent stream variants
```
joeyparrish, line 548: I honestly think it feels wrong that Packager spits out a "variant" that pairs with an audio group of itself, but I haven't spent the time digging through the spec to decide if it's required by the spec.  And Packager isn't the only one, and Player already has to handle it.  In any case, you've done well with this detail.

joeyparrish (changes requested):
That's everything!  Thanks for being patient with me.  I know it has taken me all day (my day, well into your night) to get this review completed.

streamer/m3u8_concater.py
meryacine: I think it's better to leave it a blank string as you stated. Thanks!

> It's weird, though, and maybe hacky, to put the comma and newline before the URI as part of this string. That would more logically belong to the EXTINF string.

Was fighting with line lengths T_T

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+          # An EXT-X-MAP must have a URI attribute and optionally
+          # a BYTERANGE attribute.
+          attribs = _extract_attributes(line)
+          self.content += '#EXT-X-MAP:URI=' + _quote(
+              os.path.join(period_dir, _unquote(attribs['URI'])))
+          if attribs.get('BYTERANGE'):
+            self.content += ',BYTERANGE=' + attribs['BYTERANGE']
+          self.content += '\n'
+        elif line.startswith(MediaPlaylist.HEADER_TAGS + ('#EXT-X-ENDLIST',)):
+          # Escape header and end-list tags.
```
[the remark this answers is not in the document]
meryacine: Yup

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+          pass
+        elif not line.startswith('#EXT'):
+          # Escape comments.
```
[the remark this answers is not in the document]
meryacine: Yup

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+    # for example, a codec attribute might be CODECS="avc1,ac-3,mp4a,opus",
+    # the video codec is put first then the audio codecs are put in 
+    # lexicographical order(by observation), which isn't nesseccary the same order of 
+    # #EXT-X-MEDIA in the master playlist, thus there is no solid baseground for 
+    # matching the codecs using the information in the master playlist.
+    
+    output_stream: Optional[OutputStream] = None
+    lines = self.content.split('\n')
+    for i in range(len(lines)):
+      # Don't use #EXT-X-MAP to get the codec, because HLS specs says
```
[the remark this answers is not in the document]
meryacine:
```python
def get_init_seg_file(self) -> Pipe:
    INIT_SEGMENT = {
      MediaType.AUDIO: 'audio_{language}_{channels}c_{bitrate}_{codec}_init.{format}',
      MediaType.VIDEO: 'video_{resolution_name}_{bitrate}_{codec}_init.{format}',
      MediaType.TEXT: 'text_{language}_init.{format}',
    }
    path_templ = INIT_SEGMENT[self.type].format(**self.features)
    return Pipe.create_file_pipe(path_templ, mode='w')

  def get_media_seg_file(self) -> Pipe:
    MEDIA_SEGMENT = {
      MediaType.AUDIO: 'audio_{language}_{channels}c_{bitrate}_{codec}_$Number$.{format}',
      MediaType.VIDEO: 'video_{resolution_name}_{bitrate}_{codec}_$Number$.{format}',
      MediaType.TEXT: 'text_{language}_$Number$.{format}',
    }
    path_templ = MEDIA_SEGMENT[self.type].format(**self.features)
    return Pipe.create_file_pipe(path_templ, mode='w')

  def get_single_seg_file(self) -> Pipe:
    SINGLE_SEGMENT = {
      MediaType.AUDIO: 'audio_{language}_{channels}c_{bitrate}_{codec}.{format}',
      MediaType.VIDEO: 'video_{resolution_name}_{bitrate}_{codec}.{format}',
      MediaType.TEXT: 'text_{language}.{format}',
    }
    path_templ = SINGLE_SEGMENT[self.type].format(**self.features)
    return Pipe.create_file_pipe(path_templ, mode='w')
```
```m3u8
#EXTM3U
#EXT-X-VERSION:6
## Generated with https://github.com/google/shaka-packager version c1f64e5-release
#EXT-X-TARGETDURATION:4
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-MAP:URI="video_480p_1M_h264_init.mp4"
#EXTINF:3.000,
video_480p_1M_h264_1.mp4
#EXT-X-ENDLIST
```
We use the `streams_map` to get the matching output stream object, but when creating the streams map we only added the mapping from the methods `get_single_seg_file()` and `get_media_seg_file()`, in the `segment_per_file` case, we could have added `get_init_seg_file()` instead and checked the URI in the `#EXT-X-MAP` tag, this would yield to the same result, but i was skeptic about that in some cases packager might not output an `#EXT-X-MAP` tag since it's optional according to the HLS specs.  Kept the comment as a reminder.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+    # #EXT-X-STREAM-INF tags we will have to match these codecs with each stream,
+    # for example, a codec attribute might be CODECS="avc1,ac-3,mp4a,opus",
+    # the video codec is put first then the audio codecs are put in 
+    # lexicographical order(by observation), which isn't nesseccary the same order of 
+    # #EXT-X-MEDIA in the master playlist, thus there is no solid baseground for 
+    # matching the codecs using the information in the master playlist.
+    
+    output_stream: Optional[OutputStream] = None
+    lines = self.content.split('\n')
+    for i in range(len(lines)):
```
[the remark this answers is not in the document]
meryacine: Yes.
Ok, will do.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+    AVERAGE-BANDWIDTH = (AVERAGE-BANDWIDTH)s . (DURATION)s
+                        /sum((DURATION)s)
+    """
+    
+    band, avg_band = 0, 0.0
+    for var_playlist, duration in zip(var_playlists, durations):
+      band = max(int(var_playlist.stream_info['BANDWIDTH']), band)
+      avg_band += int(var_playlist.stream_info['AVERAGE-BANDWIDTH']) * duration
+    
+    return str(band), str(math.ceil(avg_band/sum(durations)))
```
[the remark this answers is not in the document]
meryacine: I think a dictionary from the callee and an update from the caller side would be very nice.

For the return types, i try to crowd every thing that will be performed a lot or not pleasant for the eye inside the helper methods so that the concat_xxx methods don't get much bigger or uglier :(

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+      candidate = variant.language
+      candidate_split = candidate.split('-', 1)
+      candidate_base = candidate_split[0]
+      candidate_is_reg= len(candidate_split) > 1
+      # Only kick the previous best fit out when: The base of the candidate 
+      # is the same as the base of the original language AND (the base of 
+      # the best fit is not the same as the base of the original language
+      # OR the candidate is a regional variant).
+      if language_base == candidate_base:
+        if language_base != best_fit_base or candidate_is_reg:
```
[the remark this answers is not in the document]
meryacine: (regional variant > base-only variant) das ist gut.

(regional variant > another regional variant with the same base) they should be equal, but the stability isn't a concern, the user should have provided stitchable subtitles in the first place >:O.

I wouldn't say it's not stable though, it is like stable from the end, so always the regional variant at the end wins the slot.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+    return max(media_playlist.target_duration for
+               media_playlist in media_playlists)
+  
+  @staticmethod
+  def _fit_missing_lang(variant_options: List['MediaPlaylist'],
+                        language: str) -> str:
+    """Returns a substitution language for a missing language by considering
+    the languages of the given variants.
+    
+    Returns the argument `language` back only if no variants were given
```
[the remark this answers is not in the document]
meryacine: This is the argument passed to the method, it's like, if you don't provide me with variants i can choose form to replace the missing `language`, i will just return this missing language back to you.

This will only happen for when no text input in any language what so ever is present for some period, this means we will fill this period with the dummy subtitles for the period's duration to keep the subtitles of the next period(if there is) in sync.
This method will return the same missing language back and that will indicate that we failed to find a substitution for this language.

It's some fancy way to handle this though, i can just change it in the `concat_sub` method itself.

streamer/m3u8_concater.py
```diff
@@ -0,0 +1,957 @@
...
+                           codec_string in codec_strings))
+  
+  @staticmethod
+  def _next_unq_name() -> Tuple[str, str]:
+    """Returns a unique NAME and URI."""
+    
+    stream_name = 'stream_' + str(MediaPlaylist.current_stream_index)
+    MediaPlaylist.current_stream_index += 1
+    
+    return _quote(stream_name), _quote(stream_name + '.m3u8')
```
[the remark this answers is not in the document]
meryacine: That would be absolutely great.

streamer/m3u8_concater.py
meryacine: Yes, i am using this perk of the Streamer to make the matching easier, actually i had to do it similar to what you do in the Player with the audio channel matching prior to the 'channel_layout as an input feature' PR.  I was starting with a dictionary full on `None`s for every slot and start filling it appropriately with the channel count it should get for every period.

streamer/m3u8_concater.py
meryacine: Do you mean the error?
Python generates a very long float number when doing a normal float division, i think accumulating these fractions to one `#EXTINF` tag might reduce the error to some extent.

streamer/m3u8_concater.py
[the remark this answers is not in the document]
joeyparrish: Ah, now I see.  How about this, which would make more sense to me in a comment:

```
# Don't use the URIs from any tag to try to extract codec information.
# We should not rely on the exact structure of file names for this.
# Use stream_maps instead.
```

streamer/m3u8_concater.py
```diff
@@ -251,10 +242,13 @@ def _get_bandwidth(var_playlists: List['MediaPlaylist'],
       band = max(int(var_playlist.stream_info['BANDWIDTH']), band)
       avg_band += int(var_playlist.stream_info['AVERAGE-BANDWIDTH']) * duration
     
-    return str(band), str(math.ceil(avg_band/sum(durations)))
+    return {
+        'BANDWIDTH': str(band),
+        'AVERAGE-BANDWIDTH': str(math.ceil(avg_band/sum(durations)))
```
joeyparrish, line 247: Wow.  I'm really surprised to check the spec and find that these values must be integers.  Nice attention to detail!

I probably would have just written out a float, which would have worked fine in a JS player and may have broken some other player implementation.

streamer/m3u8_concater.py
joeyparrish: I mean some small accumulated rounding error here is fine.  I think you're doing the right thing, and I think it's not worth the effort to do more until we have evidence that it's necessary.

joeyparrish (changes requested):

joeyparrish:
Just back from vacation.  Will try to take another pass at review today if at all possible.

streamer/m3u8_concater.py
```diff
@@ -406,22 +407,15 @@ def non_nones(items: List[Optional[MediaPlaylist]]) -> List[MediaPlaylist]:
...
-          for _ in range(math.floor(td_count)):
-            concat_txt_playlist.content += '#EXTINF:' + str(
-                concat_txt_playlist.target_duration) + ',\n'
-            concat_txt_playlist.content += dummy
-          # Put whats left in an #EXTINF.
-          remains = round(durations[i] % concat_txt_playlist.target_duration, 3)
-          if remains:
-            concat_txt_playlist.content += '#EXTINF:' + str(remains) + ',\n'
-            concat_txt_playlist.content += dummy
+          # If the no playlist were found for this period, we create at time gap
```
joeyparrish, line 410: typo: at => a

joeyparrish (changes requested):

joeyparrish (approved):

PR #84, opened 2021-07-21
Channel count as an input feature

meryacine:
This PR removes the `channels` field from the pipeline configuration and replaces it with `channel_layouts`, a list of channel layouts to encode.
It also adds `channel_layout` field for `Input` class.

### DIFF
--- a/streamer/configuration.py
+++ b/streamer/configuration.py
@@ -561,5 +561,5 @@ class RuntimeMapKeyValidator(ValidatingType, str):
 
     if key not in cls.map_class.keys():
       raise ValueError(
-          '{} is not a valid {}, {}'.format(key, cls.map_class.__name__, cls.name()))
+          '{} is not a valid {}'.format(key, cls.name()))
 
--- a/streamer/configuration.py
+++ b/streamer/configuration.py
@@ -561,5 +561,5 @@ class RuntimeMapKeyValidator(ValidatingType, str):
 
     if key not in cls.map_class.keys():
       raise ValueError(
-          '{} is not a valid {}'.format(key, cls.map_class.__name__))
+          '{} is not a valid {}, {}'.format(key, cls.map_class.__name__, cls.name()))
 
--- a/streamer/configuration.py
+++ b/streamer/configuration.py
@@ -72,8 +72,8 @@ class MalformedField(ConfigError):
     self.reason = reason
 
   def __str__(self):
-    return 'In {}, {} field is malformed: {}, a {}'.format(
-        self.class_name, self.field_name, self.reason, self.field.get_type_name())
+    return 'In {}, {} field is malformed: {}'.format(
+        self.class_name, self.field_name, self.reason)
 
 class ConflictingFields(ConfigError):
   """An error raised when multiple fields are given and only one of them is allowed at a time."""
--- a/streamer/configuration.py
+++ b/streamer/configuration.py
@@ -72,8 +72,8 @@ class MalformedField(ConfigError):
     self.reason = reason
 
   def __str__(self):
-    return 'In {}, {} field is malformed: {}'.format(
-        self.class_name, self.field_name, self.reason)
+    return 'In {}, {} field is malformed: {}, a {}'.format(
+        self.class_name, self.field_name, self.reason, self.field.get_type_name())
 
 class ConflictingFields(ConfigError):
   """An error raised when multiple fields are given and only one of them is allowed at a time."""
--- a/streamer/input_configuration.py
+++ b/streamer/input_configuration.py
@@ -146,6 +146,9 @@ class Input(configuration.Base):
   Otherwise, it will default to 'und' (undetermined).
   """
 
+  channels = configuration.Field(int).cast()
+  """The number of audio channels to encode."""
+
   start_time = configuration.Field(str).cast()
   """The start time of the slice of the input to use.
 
@@ -217,13 +220,17 @@ class Input(configuration.Base):
         self.resolution = autodetect.get_resolution(self)
       require_field('resolution')
 
-    if self.media_type == MediaType.AUDIO or self.media_type == MediaType.TEXT:
-      # Language is required for audio and text inputs.
-      # We will attempt to auto-detect this.
+    if self.media_type == MediaType.AUDIO:
       if self.language is None:
         self.language = autodetect.get_language(self) or 'und'
 
+      if self.channels is None:
+        self.channels = autodetect.get_channel_count(self)
+      require_field('channels')
+
     if self.media_type == MediaType.TEXT:
+      if self.language is None:
+        self.language = autodetect.get_language(self) or 'und'
       # Text streams are only supported in plain file inputs.
       if self.input_type != InputType.FILE:
         reason = 'text streams are not supported in input_type "{}"'.format(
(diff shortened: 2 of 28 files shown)


meryacine:
We can also follow the `VideoResolutionName` approach and make an `AudioChannelLayoutName` and have a `channel_layout` field instead where the user can type in `stereo`, `mono`, etc..

streamer/input_configuration.py
```diff
@@ -146,6 +146,9 @@ class Input(configuration.Base):
   Otherwise, it will default to 'und' (undetermined).
   """
 
+  channels = configuration.Field(int).cast()
+  """The number of audio channels to encode."""
```
joeyparrish, line 150: It's more like "The number of channels in an audio input", right?  It's less about encoding, and more about the structure of the input.  Do I understand that right?

joeyparrish (changes requested):

streamer/input_configuration.py
meryacine: No, i feel a bit confused because of this too.
This is actually an output feature, the number of channels to encode in the final output file. sticking it with the each input(instead of the pipeline config) will allow us to give different channel counts for each input in a multi-period list.

meryacine: Another less confusing approach could be, to deal with it like resolutions, we may specify the channel counts we want as an output to be encoded with, and it's a list of channel counts rather than a single channel count and each input will be encoded with different channel counts up till its own channel count just like video resolutions.

joeyparrish: That would make sense to me.  I like having it in the input, though, with auto-detection, just like resolution and frame rate for video inputs.  If auto-detect fails, or if the input is coming from a pipe, rather than a file, we would need to be able to specify it in the input config.

So let's keep what you have (but fix the comment here on the field meaning), and then let's add a list of output channel counts in the pipeline config, just like we do for resolutions.

streamer/configuration.py
```diff
@@ -561,5 +561,5 @@ def validate(cls, key):
 
     if key not in cls.map_class.keys():
       raise ValueError(
-          '{} is not a valid {}'.format(key, cls.map_class.__name__))
+          '{} is not a valid {}, {}'.format(key, cls.map_class.__name__, cls.name()))
```
joeyparrish, line 564: From what I can tell re-reading the source, this now repeats cls.map_class.__name__, which is also printed by name() above:

"foo is not a valid VideoResolution, VideoResolution name (one of 720p, 1080p, ...)"

Should we drop cls.map_class.__name__ in favor of cls.name()?

streamer/pipeline_configuration.py
```diff
@@ -242,9 +242,13 @@ class PipelineConfig(configuration.Base):
   multiple inputs.
   """
 
-  # TODO(joeyparrish): Default to whatever is in the input.
-  channels = configuration.Field(int, default=2).cast()
-  """The number of audio channels to encode."""
+  channel_layouts = configuration.Field(
+      List[bitrate_configuration.AudioChannelLayoutName],
+      required=True).cast()
```
joeyparrish, line 247: I don't agree with making this a required field.  I think we should have a default value instead.  Perhaps `['surround', 'stereo']`.  Not sure.  What are your thoughts?

tests/tests.js
```diff
@@ -726,17 +740,17 @@ function codecTests(manifestUrl, format) {
...
       'resolutions': [],
+      'channel_layouts': ['stereo'],
       'audio_codecs': ['aac', 'opus'],
     };
     await startStreamer(inputConfigDict, pipelineConfigDict);
     await player.load(manifestUrl);
 
-    const trackList = player.getVariantTracks();
-    const audioCodecList = trackList.map(track => track.audioCodec);
+    let codecList = await getAudioAndVideoCodecs(manifestUrl)
```
joeyparrish, line 749: Style nit: Add a semicolon at the end of this statement.

tests/tests.js
```diff
@@ -726,17 +740,17 @@ function codecTests(manifestUrl, format) {
     const pipelineConfigDict = {
       'streaming_mode': 'vod',
       'resolutions': [],
+      'channel_layouts': ['stereo'],
       'audio_codecs': ['aac', 'opus'],
     };
     await startStreamer(inputConfigDict, pipelineConfigDict);
     await player.load(manifestUrl);
```
joeyparrish, line 747: No need to call player.load() if you're using getAudioAndVideoCodecs(), which will do that internally.

joeyparrish (changes requested):

streamer/configuration.py
meryacine: You read that write.
Yes i think we should drop it in favor of `cls.name()`.

streamer/pipeline_configuration.py
meryacine: I fear that someone might provide a generic bitrate config that doesn't have `stereo` in it.

My thoughts is that we can default to a list that contains only the lowest `channel_layout`. So we read the `channel_layouts` using the `BitrateConfig` class and then provide this list as a default `[AudioChannelLayout.sorted_values()[0].get_key()]`, now it will pick `stereo` by default, or, when a generic user provided bitrate config is there, it will pick the lowest of what ever the user has.

We can do the same for `resolutions` field as well.

meryacine: Too bad this is not gonna work :(,
Because `channel_layouts` and `resolutions` are class variables and are initialized at indexing time, so even before the bitrate config is provided.

meryacine: If you agree to the proposed idea above, we can monkey-patch the `default` of these fields in the `__init__` of the `PipelineConfig`.
like this.
```
self.__class__.resolutions.default=[ # type: ignore
        bitrate_configuration.VideoResolution.sorted_values()[0].get_key()]
self.__class__.channel_layouts.default=[ # type: ignore
        bitrate_configuration.AudioChannelLayout.sorted_values()[0].get_key()]
```
we need the `# type: ignore` now because since we did the `cast()` for these class attributes, mypy will think they are actual `[RuntimeMap]` while they are actually `Field` instances.

streamer/pipeline_configuration.py
```diff
@@ -307,6 +305,10 @@ class PipelineConfig(configuration.Base):
 
 
   def __init__(self, *args) -> None:
+    self.__class__.resolutions.default=[ # type: ignore
```
joeyparrish, line 308: Since resolution and channel layout have a behavior of "use every option that doesn't upscale/upmix", perhaps the best default would be the entire list.  Can you please try that and let me know how it goes?  Also, please update the docstrings for those fields to say that they default to all defined values from the bitrate config.

joeyparrish (changes requested):

joeyparrish (approved):

joeyparrish:
Looks like there are a couple of conflicts with the Windows PR.  Can you rebase/merge and fix those up please?

streamer/output_stream.py
```diff
@@ -95,38 +95,24 @@ def __init__(self,
...
-
-    assert self.layout, 'Unable to find audio layout for {} channels'.format(
-        self.channels)
+    self.layout = channel_layout
 
     # The features that will be used to generate the output filename.
     self._features = {
       'language': input.language,
-      'channels': str(self.channels),
+      'channels': str(self.layout.max_channels),
```
meryacine, line 108: Would you like having this to be `'channel_layout': self.layout.get_key()` instead.
@joeyparrish

joeyparrish (approved):

PR #86, opened 2021-07-31
Using Enum._missing_ hook to generically handle the hw: variants

meryacine:
This PR removes the `hw:` enum variants, making the enum only for the base codec variants.

If a codec is prefixed with `hw:`, it will be caught in the `_missing_` hook, we will try to know if it can be one of the base variants or not, if it is a base variant we will set the `_hw_acc` to `True` for this codec variant, if not we will fall back the the `super()._missing_` and print the appropriate error message.

### DIFF
--- a/streamer/bitrate_configuration.py
+++ b/streamer/bitrate_configuration.py
@@ -17,7 +17,7 @@ import math
 import re
 
 from . import configuration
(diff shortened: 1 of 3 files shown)


joeyparrish (approved):
This looks great!  Nicely done!

PR #87, opened 2021-08-16
Shaka Streamer Binaries

meryacine:
## Steps to build:
- run `python build_wheels.py` in the terminal.


binaries/build.py
```diff
@@ -0,0 +1,127 @@
...
+    # completed and doesn't also re-create it in the next `build` call,
+    # that's why we can't delete it either, so let's just clean it.
+    build_dir = os.path.join(self.build_lib, streamer_binaries.__name__)
+    self._clean_build_dir(build_dir)
+    return super().run()
+
+  def _clean_build_dir(self, build_dir):
+    try:
+      # To clean the directory, we can remove it and re-create it.
+      shutil.rmtree(build_dir)
```
joeyparrish, line 46: If you have time, it would be cool if build.py had constants for the release tags of Packager & ffmpeg, and would download the binaries from the GitHub repos automatically, rather than requiring us to download them manually and in advance of build.py.

streamer/controller_node.py
```diff
@@ -73,27 +75,57 @@ def start(self, output_location: str,
...
+              '  Install it with `pip install shaka-streamer-binaries`.\n'
+              '  Alternatively, use the `--use-system-binaries` option if you '
+              'want to use the system wide binaries of ffmpeg/ffprobe/packager.'
+            ) from None
+        # If we can't set the permissions for the bundled executables,
+        # we may not be able to run it as a subprocess.
+        if isinstance(ex, PermissionError):
+          raise RuntimeError(
+              'Couldn\'t set the permissions for the bundled '
+              '`shaka-streamer-binaries` executables.\n  Please run as a root.'
```
joeyparrish, line 104: Please remove the "run as root" recommendation.  If it doesn't work at runtime, we need to solve it at packaging time instead, and I'll commit to doing that.  We definitely shouldn't run all of streamer as root, and I don't want to recommend questionable security practices.

If we ultimately can't solve it at packaging time, I'll craft a message here that recommends a specific, targeted command to run as root/admin to fix the permissions.

streamer/controller_node.py
```diff
@@ -73,27 +75,57 @@ def start(self, output_location: str,
...
-      _check_version('ffprobe', ['ffprobe', '-version'], (4, 1))
-
-      # Check that Shaka Packager version is 2.4.2 or above.
-      _check_version('Shaka Packager', ['packager', '-version'], (2, 4, 2))
+      # If we are using the hermetic binaries, just check
+      # for the package's version.
+      if use_hermetic:
+        if streamer_binaries.__version__ < __version__:
+          raise VersionError(
+              'An outdated `shaka-streamer-binareis` is installed.\n'
```
joeyparrish, line 116: typo: binaries

streamer/controller_node.py
```diff
@@ -259,7 +303,8 @@ def _append_nodes_for_inputs_list(self, inputs: List[Input], output_location: st
 
     self._nodes.append(PackagerNode(self._pipeline_config,
                                     output_location,
-                                    outputs))
+                                    outputs,
+                                    self.hermetic_pacakger))
```
joeyparrish, line 307: typo: packager

Though, no joke, as I tried to type "typo" in that comment, I first accidentally typed "topy".  😄

joeyparrish (changes requested):
This looks great!  Just a few comments.

binaries/build.py
meryacine: Sure :)

binaries/download_binaries.py
```diff
@@ -0,0 +1,151 @@
...
+  # Sort the binaries based on platform.
+  package_data = {
+      # 64-bit Windows
+      'win_amd64': select_binaries(binaries_names,
+                                   name_contains=['win',]),
+      # 64-bit Linux
+      'manylinux1_x86_64': select_binaries(binaries_names,
+                                           name_contains=['linux', 'x64',]),
+      # Linux on ARM
+      'manylinux2014_aarch64': select_binaries(binaries_names,
```
meryacine, line 140: @joeyparrish As you see, it is not so reliable way to produce the yaml file we use for building the wheels xD.
Would you please name the new upcoming ARM static build with a name that has `arm64` in it, or maybe `aarch64`, and have an `x64` for linux then so we can differentiate between the two linux es.
We can also do the same for windows and macos, just for clarity.
`packager-win-x64.exe`, `packager-osx-x64.dmg` something like this.

michellezhuogg (approved):

michellezhuogg (approved):

binaries/download_binaries.py
joeyparrish: Totally.  I will update Packager's workflows to match the naming scheme I used for FFmpeg.

binaries/build.py
```diff
@@ -0,0 +1,127 @@
...
+class custom_build_py(build_py):
+  """A custom class to override the default behavior of `build_py` command."""
+
+  def run(self):
+    # Clean the build directory so the binaries that were added at the previous
+    # build don't remain in the package for the current platform we build for.
+    # This is because setuptools doesn't delete it after a build has
+    # completed and doesn't also re-create it in the next `build` call,
+    # that's why we can't delete it either, so let's just clean it.
+    build_dir = os.path.join(self.build_lib, streamer_binaries.__name__)
```
joeyparrish, line 39: I don't understand the use of `__name__` here.  Isn't `streamer_binaries.__name__` the same as `"streamer_binaries"`?

binaries/package_data.yaml
```diff
@@ -0,0 +1,14 @@
+macosx_10_9_x86_64:
+- packager-osx
+- ffprobe-osx-x64
+- ffmpeg-osx-x64
+manylinux1_x86_64:
```
joeyparrish, line 5: This file looks incomplete (no packager binary names for linux) and has no license header.  Is it generated?  Should this be in .gitignore?

binaries/platform_package_data.yaml
```diff
@@ -0,0 +1,41 @@
...
+win_amd64:
+  - ffmpeg-win-x64.exe
+  - ffprobe-win-x64.exe
+  - packager-win.exe
+
+# 64-bit Linux
+manylinux1_x86_64:
+  - ffmpeg-linux-x64
+  - ffprobe-linux-x64
+  - packager-linux
```
joeyparrish, line 29: Could you add a TODO after this to clean up the Packager release file names?

binaries/download_binaries.py
```diff
@@ -0,0 +1,152 @@
...
+      'macosx_10_9_x86_64': select_binaries(binaries_names,
+                                            name_contains=['osx',]),
+    }
+
+  # Dump `package_data` into a yaml file.
+  with open(YAML_OUTPUT_FILE, 'w') as output_file:
+    yaml.dump(package_data, output_file)
+
+
+main()
```
joeyparrish, line 152: Best practice is to have this inside a conditional:

```py
if __name__ == '__main__':
  main()
```

Then main() will not run as a side-effect of an import.  Another script or module loading this one will have to call main() explicitly, which is meant to cut down on surprises and accidents.

binaries/download_binaries.py
```diff
@@ -0,0 +1,152 @@
...
+      # Linux on ARM
+      'manylinux2014_aarch64': select_binaries(binaries_names,
+                                               name_contains=['linux', 'arm64',]),
+      # 64-bit with 10.9 SDK
+      'macosx_10_9_x86_64': select_binaries(binaries_names,
+                                            name_contains=['osx',]),
+    }
+
+  # Dump `package_data` into a yaml file.
+  with open(YAML_OUTPUT_FILE, 'w') as output_file:
```
joeyparrish, line 148: Is platform_package_data.yaml still needed in the repo?  Should both yaml files be removed from the PR, and the generated one added to .gitignore?

binaries/download_binaries.py
```diff
@@ -0,0 +1,152 @@
...
+    yaml.dump(package_data, output_file)
```
joeyparrish, line 149: Since you have modules importing modules, why pass data in a yaml file instead of a dictionary return value?

binaries/download_binaries.py
```diff
@@ -0,0 +1,152 @@
...
+    # `res_in_json[0]` will grab the latest release.
+    assets.extend(res_in_json[0]['assets'])
+
+  download_threads: List[Thread] = []
+  downloaded_files: List[str] = []
+
+  # Loop over all the assets and download them.
+  for asset in assets:
+    download_url: str = asset['browser_download_url']
+    download_threads.append(Thread(target=download_binary,
```
joeyparrish, line 92: It seems like overkill to use threads to download binaries.  I would prefer something simpler, even if it's somewhat slower.

binaries/download_binaries.py
```diff
@@ -0,0 +1,152 @@
...
+    # If we have every substring the `binary_name` so we can select it safely.
+    if all(substring.lower() in binary_name.lower() for
+           substring in name_contains):
+      selected_binaries.append(binary_name)
+  return selected_binaries
+
+
+def main():
+  # https://api.github.com/repos/{owner}/{repo}/releases
+  releases_urls = ['https://api.github.com/repos/joeyparrish/static-ffmpeg-binaries/releases',
```
joeyparrish, line 122: I think versioned download URLs could be constructed without resorting to the GitHub API.

Something like:

```py
PACKAGER_VERSION = 'v2.5.1'
FFMPEG_VERSION = 'early-trial-122`  # FIXME: update when first official release made

PACKAGER_DL_PREFIX = 'https://github.com/google/shaka-packager/releases/download/' + PACKAGER_VERSION
PACKAGER_DL = {
  'win-x64': PACKAGER_DL_PREFIX + '/packager-win.exe',
  'osx-x64': PACKAGER_DL_PREFIX + '/packager-osx',
}

# etc...
```

I think the download script could be much simpler this way.  Updates would just involve updating the tags at the top.

binaries/build.py
```diff
@@ -0,0 +1,127 @@
...
+        'Operating System :: Microsoft :: Windows',
+    ],
+    package_data={
+        # Only add the corresponding platform specific binaries
+        # to the package for the current platform_name.
+        streamer_binaries.__name__: platform_binaries,
+    },
+    # Use our custom builder.  All it does is cleaning the build directory
+    # before using it for building, as it might contain old unwanted binaries.
+    cmdclass={'build_py': custom_build_py},
```
joeyparrish, line 95: Could you consider using [`TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) instead?  Something like:

```py
with tempfile.TemporaryDirectory() as temp_dir_name:
  download_into(temp_dir_name, platform_name)
  build_binary_package(temp_dir_name, platform_name)
# Now the temporary directory and its contents have been wiped out
```

joeyparrish (changes requested):

binaries/download_binaries.py
meryacine: Sure,
I also think of combining the downloader and the builder files together.

binaries/build.py
meryacine: Thought this just might be a better way to refer to the package's name. Yes, they are the same.

binaries/build.py
meryacine: If i get this correctly, you mean to download the binaries in a temp directory and then add it to the package.
`setup` looks for `package_data` inside the package we say it resides in(in this case, inside `streamer_binaries`), i don't think we can put it some where else. If `setup` couldn't find the files we listed in `package_data` inside the package, it just skips it, doesn't search somewhere else.

meryacine: I looked up a bit about whether we are able to specify the build directory to the `bdist_wheel` command, but found we no way.  `bdist_wheel` calls `build` internally, while the `build` command has the `--build-lib` option exposed, the `bdist_wheel` doesn't.
Maybe the usage of `cmdclass` is a bit confusing.  We can simplify this to be only creating the needed build directory before the `setup` call and removing it afterwards, after the `setup` call.

meryacine: I did change it a little bit now to use the temporary directory `--bdist-dir`. This means we build directly in the same directory we package things in afterwards, `setuptools` removes this directory after building the wheel.
Kinda hacky, but i think is simpler.

joeyparrish:
Just back from vacation.  Will try to take another pass at review today if at all possible.

binaries/build_wheels.py
```diff
@@ -0,0 +1,154 @@
...
+import urllib.request
+import setuptools # type: ignore
+import setuptools.command.build_py #type: ignore
+import streamer_binaries
+
+
+# Version constants.
+# Change to download different versions.
+FFMPEG_VERSION = 'n4.4-1'
+PACKAGER_VERSION = 'v2.99.5'
```
joeyparrish, line 30: Not a real version.  :-)  v2.5.1 is out, and v2.6.0 is coming soon (< 1 week, I think).  The consistent binary suffixes are going to be in v2.6.0, so you can go ahead and pin to that version.

binaries/build_wheels.py
```diff
@@ -0,0 +1,154 @@
...
+    # 64-bit Linux
+    'manylinux1_x86_64': '-linux-x64',
+    # Linux on ARM
+    'manylinux2014_aarch64': '-linux-arm64',
+    # 64-bit with 10.9 SDK
+    'macosx_10_9_x86_64': '-osx-x64',
+  }
+
+FFMPEG_DL_PREFIX = 'https://github.com/joeyparrish/static-ffmpeg-binaries/releases/download/' + FFMPEG_VERSION
+PACKAGER_DL_PREFIX = 'https://github.com/joeyparrish/shaka-packager/releases/download/' + PACKAGER_VERSION
```
joeyparrish, line 47: Please change to google/shaka-packager.  I test some CI changes on my fork, but that is not where we will officially take binaries from.

binaries/build_wheels.py
```diff
@@ -0,0 +1,154 @@
...
+
+# The download links to each binary.  These download links
+# aren't complete, they miss the platfrom-specific postfix.
+BINARIES_DL = [
+    FFMPEG_DL_PREFIX + '/ffmpeg',
+    FFMPEG_DL_PREFIX + '/ffprobe',
+    PACKAGER_DL_PREFIX + '/packager',
+  ]
```
joeyparrish, line 55: style nit: align with `BINARIES_DL`

binaries/build_wheels.py
```diff
@@ -0,0 +1,154 @@
...
+PLATFORM_POSTFIXES = {
+    # 64-bit Windows
+    'win_amd64': '-win-x64.exe',
+    # 64-bit Linux
+    'manylinux1_x86_64': '-linux-x64',
+    # Linux on ARM
+    'manylinux2014_aarch64': '-linux-arm64',
+    # 64-bit with 10.9 SDK
+    'macosx_10_9_x86_64': '-osx-x64',
+  }
```
joeyparrish, line 44: style nit: align with `PLATFORM_POSTFIXES`

binaries/build_wheels.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Version constants.
+# Change to download different versions.
+FFMPEG_VERSION = 'n4.4-1'
+PACKAGER_VERSION = 'v2.99.5'
+
+# A map of postfixes that will be combined with the binary download links
+# to achieve a full download link.  Different postfix for each platform.
+# Extend this dictionary to add more platforms.
+PLATFORM_POSTFIXES = {
```
joeyparrish, line 35: Perhaps rename to "suffix" instead of "postfix".  I checked myself on Wikipedia, since perhaps I'm just not used to hearing postfix for the ending of a word.  What I found was the [disambiguation article](https://en.wikipedia.org/wiki/Postfix) that says:

> Postfix (linguistics), an affix which is placed after the stem of a word

But when I click [the link](https://en.wikipedia.org/wiki/Postfix_(linguistics)), it redirects to [the article for "suffix".](https://en.wikipedia.org/wiki/Suffix)

Language is hard and dumb and arbitrary, and English more than most, and **_American_** English...

binaries/build_wheels.py
```diff
@@ -0,0 +1,154 @@
...
+  def initialize_options(self):
+    return_val = super().initialize_options()
+    # Sets the `--build-lib` directory.
+    self.build_lib = custom_build_py.platform_build_lib
+    return return_val
+
+
+def build_bdist_wheel(platform_name, platform_binaries):
+  """Builds a wheel distribution for `platform_name` adding the files
+  in `platform_binaries` to it using the `package_data` parameter."""
```
joeyparrish, line 72: So the `platform_binaries` argument is how you are avoiding using a temp directory?  You are allow-listing the binaries per-platform in each call to build a wheel?

joeyparrish (changes requested):

joeyparrish:
BTW, I really like the latest changes.  Much cleaner!

binaries/build_wheels.py
meryacine: yeah, i was just testing with your version because it has the right binary names.

binaries/build_wheels.py
meryacine: ummm... not exactly, 
platform_binaries is used to tell `setuptools` which files in the package(`streamer_binaries`) to include. By default, `setuptools` will include all the `.py` files and leave the others, will not put them into the wheel, by specifying them as `package_data`, we are telling `setuptools` to ship them into the `.whl` file it is going to generate with all the `.py` files.

The build directory problem is that when `setuptools` is building the package for one platform, it dumps every thing it will ship in `build/lib`(the default), then it doesn't delete it, the next platform in the for loop will be built in the same directory `build/lib`, but some old binaries are still there, thus we have to clean this directory after compressing the wheel for each platform.

We can't set `--build-lib` from the `bdist_wheel`. `bdist_wheel` calls `build`, and `build` has the `--build-lib` option, so i set the from the `build` command directly. By setting each build directory to the platform name we avoid reusing the same build directory so we are fine :D.

binaries/build_wheels.py
meryacine: I don't get it.  Do you mean i should remove all the indentation?

meryacine: got it, my bad.

joeyparrish (approved):
I still have questions about the mechanics of building the wheel, but I approve of what you've done regardless.  Let's find time to chat so I can learn more, and I may add some more comments based on our conversation.

joeyparrish:
I spent some time playing around with the code, and I think I understand it now.  I may add a few comments later.

So as soon as you can resolve conflicts with this PR, I am ready to merge it.  Thank you for your hard work on this!  It's really exciting to see it landing.

meryacine:
@joeyparrish While testing with `--test-hermetics` enabled, the `drmTests` for DASH and HLS fails.
```
+ /home/omar/.local/lib/python3.8/site-packages/streamer_binaries/packager-linux-x64 'in=/tmp/shaka-live-05_j15_r/5953b968-e71f-47fd-ba32-a663cc91e2b8,stream=video,init_segment=output_files/video_144p_108k_h264_init.mp4,segment_template=output_files/video_144p_108k_h264_$Number$.mp4' --quiet --segment_duration 4 --generate_static_live_mpd --mpd_output output_files/dash.mpd --hls_playlist_type VOD --hls_master_playlist_output output_files/hls.m3u8 --enable_widevine_encryption --key_server_url https://license.uat.widevine.com/cenc/getcontentkey/widevine_test --content_id EC71357493E0C07F58469229C8E31B38 --signer widevine_test --aes_signing_key 1ae8ccd0e7985cc0b6203a55855a1034afc252980e970ca90e5202689f947ab9 --aes_signing_iv d58ce954203b7c9a9a9d467f59839249 --protection_scheme cenc --clear_lead 0
[0904/092728:ERROR:proto_json_util.cc(34)] Failed to parse from JSON:  error: Unexpected end of string. Expected a value.

^
[0904/092728:ERROR:widevine_key_source.cc(409)] Failed to convert JSON to proto: 
[0904/092728:ERROR:packager_util.cc(80)] Widevine encryption key source failed to fetch keys: 13 (SERVER_ERROR): Failed to extract encryption key from ''.
Chrome 92.0.4515 (Linux 0.0.0) Shaka Streamer has widevine encryption enabled (hls) FAILED
        Error
            at <Jasmine>
            at UserContext.<anonymous> (tests.js:544:49)
        shaka.util.Error {
          "severity": 2,
          "category": 1,
          "code": 1001,
          "data": [
            "http://localhost:5000/output_files/hls.m3u8",
            500,
            "Some processes exited with non-zero exit codes",
            {
              "content-length": "46",
              "content-type": "text/html; charset=utf-8, text/plain"
            },
            0
          ],
          "handled": false
        } thrown
        Error: Failed to close Shaka Streamer
            at stopStreamer (tests.js:57:11)
            at async UserContext.<anonymous> (tests.js:80:5)
```
and 
```
+ /home/omar/.local/lib/python3.8/site-packages/streamer_binaries/packager-linux-x64 'in=/tmp/shaka-live-0xyurn9g/568c9820-33ed-4afd-a1c6-a998b03e169d,stream=video,init_segment=output_files/video_144p_108k_h264_init.mp4,segment_template=output_files/video_144p_108k_h264_$Number$.mp4' --quiet --segment_duration 4 --generate_static_live_mpd --mpd_output output_files/dash.mpd --hls_playlist_type VOD --hls_master_playlist_output output_files/hls.m3u8 --enable_widevine_encryption --key_server_url https://license.uat.widevine.com/cenc/getcontentkey/widevine_test --content_id EC71357493E0C07F58469229C8E31B38 --signer widevine_test --aes_signing_key 1ae8ccd0e7985cc0b6203a55855a1034afc252980e970ca90e5202689f947ab9 --aes_signing_iv d58ce954203b7c9a9a9d467f59839249 --protection_scheme cenc --clear_lead 0
[0904/092732:ERROR:proto_json_util.cc(34)] Failed to parse from JSON:  error: Unexpected end of string. Expected a value.

^
[0904/092732:ERROR:widevine_key_source.cc(409)] Failed to convert JSON to proto: 
[0904/092732:ERROR:packager_util.cc(80)] Widevine encryption key source failed to fetch keys: 13 (SERVER_ERROR): Failed to extract encryption key from ''.
Chrome 92.0.4515 (Linux 0.0.0) Shaka Streamer has widevine encryption enabled (dash) FAILED
        Error
            at <Jasmine>
            at UserContext.<anonymous> (tests.js:544:49)
        shaka.util.Error {
          "severity": 2,
          "category": 1,
          "code": 1001,
          "data": [
            "http://localhost:5000/output_files/dash.mpd",
            500,
            "Some processes exited with non-zero exit codes",
            {
              "content-length": "46",
              "content-type": "text/html; charset=utf-8, text/plain"
            },
            0
          ],
          "handled": false
        } thrown
        Error: Failed to close Shaka Streamer
            at stopStreamer (tests.js:57:11)
            at async UserContext.<anonymous> (tests.js:80:5)
```
I think it's a packager issue in the new static-build, it passes these tests with the old binaries i have in `PATH`.

meryacine:
@joeyparrish Please don't merge this yet.

joeyparrish:
> I think it's a packager issue in the new static-build, it passes these tests with the old binaries i have in PATH.

I found the issue (in Packager) and have a fix in review.  This will be followed by the official v2.6 release of Packager.

Thanks for catching this!

binaries/auto_build_setup.py
```diff
@@ -0,0 +1,54 @@
...
+import setuptools # type: ignore
+
+import streamer_binaries
+
+# The last argument passed MUST be a list of binary file names
+# that we wish include in the build.
+# Parse this argument into a list.
+# This argument will look something like this:
+# 	['binary_1', 'binary_2', 'binary_3', 'binary_4']
+platform_binaries = ast.literal_eval(sys.argv[-1])
```
joeyparrish, line 28: I like this latest change overall, but it seems hacky to encode a python list into a single arg, then parse it using "ast", when we can naturally pass multiple args.  I would prefer we use a more typical command-line convention like "--" to separate the setuptools args from the list of binaries.

For example, the caller would look like:

```py
args = [
  'foo',
  'bar',
  '--py3',
  '--',
] + binary_paths
```

And here we would do something like:

```py
separator_index = sys.argv.index('--')
platform_binaries = sys.argv[separator_index+1:]
sys.argv = sys.argv[:separator_index]
```

What do you think?

binaries/setup.py
```diff
@@ -12,24 +12,19 @@
...
-platform_binaries = ast.literal_eval(sys.argv[-1])
-assert isinstance(platform_binaries, list)
-
-# The rest are the command line arguments given to setup() function.
-sys.argv = sys.argv[:-1]
+try:
+  separator_index = sys.argv.index('--')
+except ValueError:
+  # If '--' is not passed, `platform_binaries` will be an empty list.
+  separator_index = len(sys.argv)
```
joeyparrish, line 25: I think that this should be an error.  If there aren't any platform binaries, something is definitely wrong.

You could just let it raise the exception, or you could raise a RuntimeError with an explanatory message (like "don't run setup.py directly")

meryacine: My intent was to have it usable as a normal setup.py.
But your point makes sense, it's all about the binaries.

joeyparrish (approved):
Looks good, but I will wait for your confirmation before I merge it.  Thanks!

meryacine:
@joeyparrish I think it'a all fine now.  I have no more updates.
Thanks.

PR #91, opened 2021-09-08
fix(HLS concatenation): Use `posixpath` for any path joining written in `self.content`

meryacine:
## Changes:
- Namespacing the `BaseURL` tag that we insert under `"urn:mpeg:dash:schema:mpd:2011"`.  It would work fine either way(with or without the namespace) since we register this namespace we insert as the default namespace a couple of line later.  But this change was to fix a potential error that could occur if we stop registering `default_dash_namespace`(==`"urn:mpeg:dash:schema:mpd:2011"`) to be the default one.
- Using `posixpath` module for joining the paths to the media segments.  Since we write these paths in media playlist files at the end, they should all be using a forward slash as a separator, which is not the case for windows using `os.path`.  This wasn't breaking in Shaka Player for some reason, but i reckon it might break in other HLS players.


joeyparrish (approved):
Nice!

PR #94, opened 2021-09-09
fix(errors): probe whether the input exists or not

### LINKED ISSUE #89: Need clear error message if audio input contains no audio
If an input with `type: audio` contains no audio, we get an error complaining about missing channel layout info.  Instead, we should have a clear and useful error that the input audio stream could not be found.  This should be generalized to all input types.

meryacine:
Probing with `'stream=index'` should return the `track_num` if it existed, and an empty string if the input did not exist.
fixes #89


joeyparrish (approved):
Nice work!

PR #95, opened 2021-09-09
webcam input for Windows and Microphone support

meryacine:
Adds Windows webcam support using
>`dshow` (DirectShow) input device which is the preferred option for Windows users.

Quoted from here https://trac.ffmpeg.org/wiki/Capture/Webcam

A new input type was added 'microphone' to add support for microphone devices.
On Linux, we use PulseAudio to capture audio input.
On macOS, AVFoundation is used -> this was not tested.
On Windows, Directshow is used for both microphone and webcam support.

### DIFF
--- a/streamer/input_configuration.py
+++ b/streamer/input_configuration.py
@@ -304,6 +304,7 @@ class Input(configuration.Base):
         ]
       elif platform.system() == 'Windows':
         return [
+            # Treat the input as a directshow input device.
             '-f', 'dshow',
         ]
       else:
--- a/streamer/input_configuration.py
+++ b/streamer/input_configuration.py
@@ -21,6 +21,21 @@ from . import configuration
 from typing import List, Dict, Any, Optional
 
 
+class InputNotFound(configuration.ConfigError):
+  """An error raised when an input stream is not found."""
+
+  def __init__(self, input):
+    super().__init__(input.__class__, 'track_num',
+                     getattr(input.__class__, 'track_num'))
+    self.input = input
+
+  def __str__(self):
+    return ('In {}, {} track #{} was'
+            ' not found in "{}"').format(self.class_name,
+                                         self.input.media_type.value,
+                                         self.input.track_num,
+                                         self.input.name)
+
 class InputType(enum.Enum):
   FILE = 'file'
   """A track from a file.  Usable only with VOD."""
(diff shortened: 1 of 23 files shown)


streamer/input_configuration.py
```diff
@@ -40,6 +40,15 @@ class InputType(enum.Enum):
   Does not support media_type of 'text'.
   """
 
+  MICROPHONE = 'microphone'
+  """A microphone device.  Usable only with live.
+
+  The device path should given in the name field.  For example, on Linux, this
+  might be "default".
+
+  Does not support media_type of 'text'.
```
joeyparrish, line 49: Only supports media_type of "audio", I would presume?

joeyparrish (changes requested):
Looks good otherwise.  Thanks!

joeyparrish (approved):

streamer/input_configuration.py
```diff
@@ -37,7 +37,16 @@ class InputType(enum.Enum):
   The device path should be given in the name field.  For example, on Linux,
   this might be /dev/video0.
 
-  Does not support media_type of 'text'.
+  Only supports media_type of 'video'.
```
joeyparrish, line 40: Are you sure this change was correct?  I suspect you could have a webcam with a built-in mic and get audio from it.

meryacine: I didn't know that webcams may contain mics O_O.
But it think it depends more on the input device we are using for this platform. for example `video4linux2` can't be used an audio input device, while `dshow` would work for both audio and video, so we may have an input like this on Windows:
```yaml
inputs:
    - name: "video=Integrated Camera"
      input_type: microphone
      media_type: video
```
and it would still work as if it were a webcam video input and not a microphone.
and we may use webcam input type to capture a microphone device too
```yaml
inputs:
    - name: "audio=Microphone Name"
      input_type: webcam
      media_type: audio
```
this happens on Windows and i think with the `avfoundation` on macOS as well.
Although a microphone input device a using webcam input type might work, i don't think this is how the input config should be for it. it should be using a microphone input type when capturing a mic device.

I believe in the case you mentioned, a webcam device that has a built-in mic, the mic would still show up as an audio input device that our args we provide for microphone devices(`PulseAudio`, `AVFoundation`, `dshow`) would still work for, so why not declare this input device as a mic in the first place.

joeyparrish (approved):

PR #96, opened 2021-09-20
fix(external command): Create the new shell process in a new group

### LINKED ISSUE #46: Orphaned subprocesses after shutting down with CTRL+C
Using the latest source from `master`, I find that shutting down the [Shaka Player History live stream](https://shaka-player-demo.appspot.com/demo/#asset=https://storage.googleapis.com/shaka-live-assets/player-source.m3u8) with CTRL+C causes our external command input ([gource](https://gource.io/)) and gsutil both to be orphaned instead of shut down properly.

meryacine:
When launching a process as a shell, what happens is that `Popen` creates a shell then the shell forks itself and calls `exev` system call to run the program we intended to run in the shell.

```
Popen
  |
  V
shell -fork-> another-shell
  |                 |
  V                exev
                    |
                    V
              external-command
```
We have access to the shell it self from `self._process` which we actually don't need. We need the pid of `external-command`  instead.

This PR launches the spawned shell in a new group so that we can just kill the group all together when `stop` is called.

fixes #46

### DIFF
--- a/binaries/build_wheels.py
+++ b/binaries/build_wheels.py
@@ -54,7 +54,7 @@ BINARIES_DL = [
     PACKAGER_DL_PREFIX + '/packager',
 ]
 
-BINARIES_ROOT_DIR = os.path.dirname(__file__)
+BINARIES_ROOT_DIR = os.path.abspath(os.path.dirname(__file__))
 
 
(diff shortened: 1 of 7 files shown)


joeyparrish:
Thank you!

PR #103, opened 2021-11-12
feat(cloud): Upload through HTTP proxy node

### LINKED ISSUE #47: Replace gsutil-based CloudNode with local authentication proxies and HTTP output support

mariocynicys:
Changes:
 - Disable multiperiod when using cloud upload (for now)
 - Use default credentials from GCS or S3 Python libraries to upload
 - Remove gsutil dependency and use threaded HTTP server, without temp files

Closes #47


streamer/proxy_node.py
```diff
@@ -0,0 +1,365 @@
...
+    # Store the manifest files locally in `self._temp_dir`.
+    if self._temp_dir is not None and self.path.endswith(('.mpd', '.m3u8')):
+      body = self._write_body_and_get_file_io()
+      if content_length == None:
+        # We need to re-chunk it again as we wrote it as a whole
+        # to the filesystem.
+        encode_chunked = True
+    else:
+      content_length = content_length and int(content_length)
+      body = RequestBodyAsFileIO(self.rfile, content_length)
```
joeyparrish, line 86: I'm surprised that any of this is needed: either the file IO wrapper or the temporary writing of chunks to the FS.  Also, writing chunks to the FS means the output is no longer being streamed to chunk-capable clouds like Amazon's.  Ideally, we would not buffer a whole segment before transmitting it.

I'm going to experiment with this a bit locally and see what I can do.

streamer/proxy_node.py
```diff
@@ -0,0 +1,365 @@
...
+    url += self._params_and_querystring
+
+    self._conn.request('PUT', url, body, headers, encode_chunked=encode_chunked)
+    res = self._conn.getresponse()
+    # Disable response logging.
+    self.log_request = lambda _: None
+    # Respond to Shaka Packager with the response we got.
+    self.send_response(res.status)
+    self.end_headers()
+    # self.wfile.write(res.read())
```
joeyparrish, line 101: Cruft; please drop

streamer/proxy_node.py
```diff
@@ -0,0 +1,365 @@
...
+      if res.status == OK:
+        res_body = json.loads(res.read())
+        # Update the Authorization header that the request factory has.
+        auth = res_body['token_type'] + ' ' + res_body['access_token']
+        self.RequestHandlersFactory.update_headers(Authorization=auth)
+      else:
+        print("Couldn't refresh access token. ErrCode: {}, ErrMst: {!r}".format(
+            res.status, res.read()))
+    else:
+      print("Non sufficient info provided to refresh the access token.")
```
joeyparrish, line 284: English nit: "Insufficient"

joeyparrish (changes requested):
I'm going to play with this locally for a bit.  I might be able to make some concrete suggestions or contribute to your branch.

streamer/proxy_node.py
meryacine: In here we will write only the manifests to the FS so we can perform period concatenation on them (note: the period concatenation is carried out by the streamer and not packager, if the manifests are pushed to the remote url straight away, we won't be able to perform period concatenation). We won't write media segments to the FS though.

But, re-looking at this, I think I over complicated it (to avoid introducing new dependencies, I believe). I will look into a package that supports streaming to carry this out (probably [`requests`](https://pypi.org/project/requests/)).

shaka-streamer
```diff
@@ -60,15 +60,16 @@ def main():
...
-  parser.add_argument('-c', '--cloud-url',
-                      default=None,
-                      help='The Google Cloud Storage or Amazon S3 URL to ' +
-                           'upload to.  (Starts with gs:// or s3://)')
   parser.add_argument('-o', '--output',
                       default='output_files',
                       help='The output folder to write files to, or an HTTP ' +
                            'or HTTPS URL where files will be PUT.' +
                            'Used even if uploading to cloud storage.')
+  parser.add_argument('-H', '--add-header',
```
joeyparrish, line 68: I found it difficult to figure out how to get the necessary headers for a refresh token.  I can enable oauth2 in GCP, get a client-id and client-secret, but I don't understand how to get an initial refresh-token in an easy way.

Here are some ideas:

1. Streamer could get the initial refresh token for me, just based on client-id and client-secret
2. There would be docs (maybe in the examples of docs/source/overview.rst) showing how to get refresh-token
3. We could use a library like [google-auth](https://google-auth.readthedocs.io/en/master/user-guide.html) to handle the authentication (haven't found the s3 equivalent yet)

I found a quick and easy way to make it work with this code, with gcloud already installed and authenticated, but I don't think it's a "best practice":

```
  -H "Authorization=Bearer $(gcloud auth print-access-token)" \
  -o gs://my_gcs_bucket/folder/
```

It seems like this might be equivalent to using "application default credentials" in google-auth:

```py
import google.auth
credentials, project = google.auth.default()
```

meryacine: It is indeed not user friendly and needs a rework. I would go for the option of using a client id+secret, the third option is ambiguous to me right now but I will look into it.

BTW, I was getting refresh tokens from [google's oauth 2.0 playground](https://developers.google.com/oauthplayground/).

joeyparrish:
My `-H "Authorization=Bearer $(gcloud auth print-access-token)"` hack failed after about one hour of running.  I started getting HTTP 429 responses from Google Cloud, which makes me think it wasn't an issue with the token itself, but something more fundamental we will need to resolve.

[Google's docs](https://cloud.google.com/storage/docs/request-rate) say:

> If you run into any issues such as increased latency or error rates, pause your ramp-up or reduce the request rate temporarily in order to give Cloud Storage more time to scale your bucket. You should use [exponential backoff](https://cloud.google.com/storage/docs/retry-strategy#exponential-backoff) to retry your requests when:
>
> Receiving errors with 408 and 429 response codes.
> Receiving errors with 5xx response codes.

joeyparrish:
I'm going to revive this and push changes into your branch so you are properly credited for the PR.  Changes I'm planning for this PR:

1. Go back to using the cloud URL in -c for compatibility for now
2. Revert extra headers feature for now (use case unclear for now)
3. Remove upload to HTTP without gs:// or s3:// (use case unclear for now)
4. Ignore -o if -c given (no need for local files if uploading to cloud)
5. Revert changes to periodconcat_node and mark it as incompatible with cloud upload (already can't be used for live, helps avoid local temp files)
6. Remove use_local_proxy from pipeline config, implied by -c
7. Pipeline directly from incoming request to outgoing request in proxy_node
8. Try to handle authentication via default credentials in gcs and s3 python modules

I'll follow up later with new changes to:

1. Allow cloud URLs to be placed into -o
2. Make -c a deprecated alias to -o (overrides -o if both given)
3. Add command line arg to pass initial access tokens for oauth via file, maybe also environment variable

joeyparrish:
I've started by rolling up your original commits, rebasing them, and force-pushing.

joeyparrish:
@mariocynicys, still a WIP since I haven't finished the S3 upload yet, but it's working well enough for GCS that I'm going to deploy it to my lab to revive our live stream.  Without this, my live stream was running into errors where gsutil would leave files behind in /tmp and fill up the disk.

Please take a look and give me your thoughts.

joeyparrish:
Okay, S3 is working now, too.

joeyparrish:
This is working in our lab deployment now!

joeyparrish (approved):

jspizziri:
@joeyparrish now that 1.0.0 is out we started using it and discovered that [HTTP was removed](https://github.com/shaka-project/shaka-streamer/pull/103#issuecomment-2428271409)

> Remove upload to HTTP without gs:// or s3:// (use case unclear for now)

This was actually how we were using the streamer, as we use storage that is neither S3 or GCS. If you're doing that, the only option was HTTP. If there was some kind of storage adapter that would be nice and we'd write our own adapter, however in the absence of that HTTP was our only option.

Any possibility to get it added back?

joeyparrish:
Yes, definitely.  What are your requirements?

Do you need specific headers or auth tokens, or just for Packager to write output directly to an HTTP(S) URL?

joeyparrish:
@jspizziri, I filed #210 to discuss your requirements for HTTP upload.  Please let us know what you need!

PR #117, opened 2022-03-03
chore: Update URLs after moving projects

joeyparrish:


theodab (commented):
Is this really all of them? For example, a grep for "@see https://github.com/google/" returns 9 results. Some of those are for the closure compiler, which isn't moving, but it also includes stuff like a link to a Shaka Player issue in `externs/cmcd.js`.

theodab (approved):

PR #120, opened 2022-03-09
fix: only use the first line of ffprobe's output

### LINKED ISSUE #119: TS input autodetection fails
Hello,

Looks like the steamer cannot use ts-files as input. On first test the probe failed getting/setting the framerate. I then defined the fps on input config, but then it failed to get other video properties. If I manually convert the file to mp4, it can be be used fine.

I am trying to use ffmpeg filters to burn the dvb-subtitles into the video  and therefore I need to use ts  as input.

meryacine:
With some container formats (such as `ts`), ffprobe reports the output twice with newline characters in between. This fix is to take into consideration only the first line of ffprobe's output.

Closes #119


joeyparrish (approved):
Thank you!

PR #145, opened 2023-09-12
fix: Do not encode in mono by default

joeyparrish:
Encoding in mono feels like it is a waste for most stereo or surround content.  Let's only do stereo and surround by default.

### DIFF
--- a/streamer/pipeline_configuration.py
+++ b/streamer/pipeline_configuration.py
@@ -256,13 +256,13 @@ class PipelineConfig(configuration.Base):
   """
 
   channel_layouts = configuration.Field(
-      List[bitrate_configuration.AudioChannelLayoutName]).cast()
+      List[bitrate_configuration.AudioChannelLayoutName],
+      default=['stereo', 'surround']).cast()
   """A list of channel layouts to encode.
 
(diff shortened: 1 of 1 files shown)


mariocynicys:
> Let's only do stereo and surround by default.

What if the input's channel layout is already mono? We won't be able to up-mix it!

Also, hard coding the layouts like that will cause the streamer to fail if the user provides a bitrate config file that doesn't have the labels `stereo` or `surround`:

```yaml
audio_channel_layouts:
  mono:
    max_channels: 1
    bitrates:
      aac: '64k'
      opus: '32k'
  two: # not stereo
    max_channels: 2
    bitrates:
      aac: '128k'
      opus: '64k'
  six: # not surround
    max_channels: 6
    bitrates:
      aac: '192k'
      opus: '96k'
```

<hr>

I think the way it is right now (using all the bitrate config layouts) is the most general and robust. And it's only triggered if we didn't specify `channel_layouts` in the pipeline config anyways.

joeyparrish:
I tend to think if the defaults aren't good for me, maybe they aren't good for others... But perhaps I'm overindexing on myself on this one.  Thanks for the feedback!

PR #154, opened 2024-08-29
test: Fix multiperiod tests

joeyparrish:
Multiperiod tests make requests for content in subfolders of the output folder.  However, the werkzeug utility "secure_filename" ruins this by replacing "/" with "_".

secure_filename() was used to prevent breakout attempts (requests for "../../../../../../../boot/vmlinuz" for example), but was intended by its authors for use in file uploads where a single filename was given, not a full path.  So we replace this with a simple check for ".." in the path.

This went unnoticed since the introduction of multi-period support because there was no GitHub Actions workflow to run the tests.


JulianDomingo (approved):
Thanks for the explanation in the conv. thread!

PR #161, opened 2024-10-11
feat: Expand binary platforms and hardware encoding

joeyparrish:
- Update to the latest Shaka Packager (v3.2.0) and FFmpeg (n7.1) releases
 - Add support for macOS arm64
 - Add support for hardware encoding on Ubuntu Linux 22.04 and 24.04
 - Adjust AV1 encoding options to match new FFmpeg release

### DIFF
--- a/binaries/build_wheels.py
+++ b/binaries/build_wheels.py
@@ -26,33 +26,41 @@ import streamer_binaries
 
 # Version constants.
 # Change to download different versions.
-FFMPEG_VERSION = 'n4.4-2'
-PACKAGER_VERSION = 'v2.6.1'
+FFMPEG_VERSION = 'n7.1-1'
+PACKAGER_VERSION = 'v3.2.0'
 
 # A map of suffixes that will be combined with the binary download links
 # to achieve a full download link.  Different suffix for each platform.
 # Extend this dictionary to add more platforms.
 PLATFORM_SUFFIXES = {
-    # 64-bit Windows
-    'win_amd64': '-win-x64.exe',
-    # 64-bit Linux
-    'manylinux1_x86_64': '-linux-x64',
-    # Linux on ARM
+    # Linux x64
+    'manylinux2014_x86_64': '-linux-x64',
+    # Linux arm64
     'manylinux2014_aarch64': '-linux-arm64',
-    # 64-bit with 10.9 SDK
+    # macOS x64 with 10.9 SDK
     'macosx_10_9_x86_64': '-osx-x64',
+    # macOS arm64 with 10.9 SDK
+    'macosx_10_9_arm64': '-osx-arm64',
+    # Windows x64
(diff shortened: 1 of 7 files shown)


binaries/setup.py
```diff
@@ -41,5 +41,10 @@
   package_data={
       # Only add the corresponding platform specific binaries to the wheel.
       streamer_binaries.__name__: platform_binaries,
-  }
+  },
+  install_requires=[
+      # This is only used for Linux, and only supports Linux, but can be
+      # installed on all platforms and won't crash if imported or run.
+      'distro',
+  ],
```
mariocynicys, line 49: can't we do something like this: https://github.com/shaka-project/shaka-streamer/blob/09df9fbbef53fdf8430cd898ea671bf81f9fd125/setup.py#L34

so we don't have to install it on other platforms.

binaries/streamer_binaries/__init__.py
```diff
@@ -31,3 +38,11 @@
 packager = os.path.join(_dir_path, 'packager-{}-{}'.format(_os, _cpu))
 """The path to the installed Shaka Packager binary."""
 
+# Special overrides for Ubuntu builds with hardware encoding support.
+# These are not static binaries, and so they must be matched to the distro.
+if _os == 'linux':
+  if distro.id() == 'ubuntu':
+    if distro.version() in _ubuntu_versions_with_hw_encoders:
+      suffix = '-ubuntu-{}'.format(distro.version())
```
mariocynicys, line 46: nit: we should be able to add them like this:
```python
'-ubuntu-' + distro.version()
```

binaries/build_wheels.py
```diff
@@ -26,33 +26,41 @@
...
+# suffix (Linux only).
+FFMPEG_BINARIES_DL = [
     FFMPEG_DL_PREFIX + '/ffmpeg',
     FFMPEG_DL_PREFIX + '/ffprobe',
+]
+PACKAGER_BINARIES_DL = [
     PACKAGER_DL_PREFIX + '/packager',
 ]
+UBUNTU_SUFFIXES = map(
+    lambda version: '-ubuntu-{}'.format(version),
```
mariocynicys, line 64: nit: the literal `'-ubuntu-'` is used two times in here. maybe we can fit it inside `_ubuntu_versions_with_hw_encoders`.

joeyparrish: That would complicate `if distro.version() in _ubuntu_versions_with_hw_encoders:`

binaries/setup.py
joeyparrish: Done

binaries/streamer_binaries/__init__.py
joeyparrish: Done

mariocynicys (approved):

PR #165, opened 2024-10-24
fix(cloud): Retry on all failures

joeyparrish:
Uses basic retry settings for GCS and S3.

If we do not pass these params, GCS would normally only retry if certain params were passed so that it knew the operation was definitely idempotent.  We know our use case, so we know we can safely retry.

There will never be two versions of the same segment file, or two concurrent writes to the MPD, for example.  So both uploads and deletes can be safely retried.

### DIFF
--- a/streamer/proxy_node.py
+++ b/streamer/proxy_node.py
@@ -60,6 +60,7 @@ except:
 # Optional: To support S3, import AWS's boto3 library.
 try:
   import boto3  # type: ignore
+  import botocore.config  # type: ignore
   SUPPORTED_PROTOCOLS.append('s3')
 except:
   pass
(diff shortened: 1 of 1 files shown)


joeyparrish:
Still verifying this fix.  It takes a long time for these failures to show up, but they were being treated as fatal before.

joeyparrish:
This appears to be working.  I hadn't made it past 6 hours with the previous revision, and we just reached 7.5 without errors with this one.

joeyparrish:
This is definitely working.  In combination with some other changes, also in Packager, I've made it through several loops in my live stream now without errors.  It has been running for the last 13.5 hours.  I think with this, we have the most stable and performant cloud upload feature we've ever had.

mariocynicys (approved):

PR #166, opened 2024-10-24
fix: Fix typing of log_request

joeyparrish:
This mistake slipped through because we are not yet validating every PR with mypy.

### DIFF
--- a/streamer/proxy_node.py
+++ b/streamer/proxy_node.py
@@ -21,7 +21,7 @@ import traceback
 import urllib.parse
 
 from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
-from typing import Any, BinaryIO, Optional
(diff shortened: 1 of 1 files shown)


mariocynicys (approved):

mariocynicys:
> This mistake slipped through because we are not yet validating every PR with mypy.

But I see we do lint check the code in the end-to-end tests: https://github.com/shaka-project/shaka-streamer/blob/aa1f4e793750573079a71bdacd11a64cf43d5730/run_end_to_end_tests.py#L310-L314

joeyparrish:
> > This mistake slipped through because we are not yet validating every PR with mypy.
> 
> But I see we do lint check the code in the end-to-end tests:

Yes, and soon I intend to revive the workflows that run those tests.  I am also going to build release workflows, like those we have for most of the other shaka-project repos.

PR #169, opened 2024-10-28
test: Fix mypy types

joeyparrish:
On some versions, mypy thinks the file handle from the web server is BinaryIO, in others, BufferedIOBase.  We don't care which it is, as both have the methods we care about.

I tried the IOBase class from "io", but I get failures on certain devices.  This is the best solution I've found so far.

### DIFF
--- a/streamer/proxy_node.py
+++ b/streamer/proxy_node.py
@@ -229,7 +229,7 @@ class GCSHandler(RequestHandlerBase):
                rate_limiter: RateLimiter, *args, **kwargs) -> None:
     self._bucket: google.cloud.storage.Bucket = bucket
     self._base_path: str = base_path
-    self._chunked_output: Optional[Union[BinaryIO, BufferedIOBase]] = None
+    self._chunked_output: Optional[BinaryIO] = None
 
     # The HTTP server passes *args and *kwargs that we need to pass along, but
     # don't otherwise care about.  This must happen last, or somehow our
--- a/streamer/proxy_node.py
+++ b/streamer/proxy_node.py
@@ -21,6 +21,7 @@ import traceback
 import urllib.parse
 
 from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
+from io import BufferedIOBase
 from typing import Any, BinaryIO, Optional, Union
 
 from streamer.node_base import ProcessStatus, ThreadedNodeBase
@@ -198,7 +199,7 @@ class RequestHandlerBase(BaseHTTPRequestHandler):
 
   @abc.abstractmethod
   def handle_non_chunked(self, path: str, length: int,
-                         file: BinaryIO) -> None:
+                         file: Union[BinaryIO, BufferedIOBase]) -> None:
     """Write the non-chunked data stream from |file| to the destination."""
     pass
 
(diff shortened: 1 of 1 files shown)


streamer/proxy_node.py
```diff
@@ -198,7 +199,7 @@ def do_DELETE(self) -> None:
 
   @abc.abstractmethod
   def handle_non_chunked(self, path: str, length: int,
-                         file: BinaryIO) -> None:
+                         file: Union[BinaryIO, BufferedIOBase]) -> None:
```
mariocynicys, line 202: i think shutting down mypy at https://github.com/joeyparrish/shaka-streamer/blob/63aae3f9c304ebded7bdd260cc3af92e0e26dd32/streamer/proxy_node.py#L256 would be simpler/clearer.

joeyparrish: I don't think that addresses the issue.  Here's an example of the error I'm getting from CI when I turn it back on:

```
The type checker found the following errors: 
streamer/proxy_node.py:161: error: Argument 3 to "handle_non_chunked" of "RequestHandlerBase" has incompatible type "BufferedIOBase"; expected "BinaryIO"  [arg-type]
Found 1 error in 1 file (checked 19 source files)
```

https://github.com/joeyparrish/shaka-streamer/actions/runs/11561541232/job/32180782250

joeyparrish: This is slightly complicated by the fact that my laptop seems to think it's BinaryIO, but the GitHub VMs call it BufferedIOBase.  The two seem to be unrelated in terms of inheritance.

mariocynicys: ah sorry i didn't test this. the `type: ignore` should be where the error is.

streamer/proxy_node.py
```diff
@@ -228,15 +229,15 @@ def __init__(self, bucket: Any, base_path: str,
                rate_limiter: RateLimiter, *args, **kwargs) -> None:
     self._bucket: google.cloud.storage.Bucket = bucket
     self._base_path: str = base_path
-    self._chunked_output: Optional[BinaryIO] = None
+    self._chunked_output: Optional[Union[BinaryIO, BufferedIOBase]] = None
```
mariocynicys, line 232: i suggested the prev solution because i thought the issue stems from here. seems this is not that case though, why is this needed?

joeyparrish: I did a search-and-replace on BinaryIO.  I'll double-check if this is needed.

joeyparrish: Yeah, this one is irrelevant.  I'll revert it.

joeyparrish: Done

streamer/proxy_node.py
mariocynicys: https://github.com/mariocynicys/shaka-streamer/actions/runs/11562438825/job/32183606442#step:6:8, yup that works, should we use it instead?

looks like there are other errors though (in lists and dicts typing) with different python versions https://github.com/mariocynicys/shaka-streamer/actions/runs/11562438825/job/32183606011

joeyparrish: I'm going to update the python versions we support, set the minimum to 3.9 going forward.

I would rather not use "type: ignore" where I can avoid it, because "A|B" is better than "whatever".  If we make a typo later, I don't want to "type: ignore" to keep us from finding out until runtime.

joeyparrish:
This PR is necessary to enable the PR testing workflow.

joeyparrish:
This is the last remaining PR now to enable the tests and have them pass.  I have a test workflow PR ready.  After that, I have PRs queued up for the release workflow.

mariocynicys (approved):
lgtm

PR #171, opened 2024-10-28
feat!: Bump minimum Python version to 3.9

joeyparrish:
Also reverts old fixes specific to Python 3.6.

Python 3.9 is the last living release, supported until EOY25.


joeyparrish:
This PR is necessary to enable the PR testing workflow.

PR #184, opened 2024-11-04
ci: Automated release workflow

joeyparrish:
To enable automated releases, this:

 - reformats the change log to match what release-please wants/needs
 - rolls back version numbers to match the latest release and tags them, so release-please can update them based on conventional commits
 - fixes a warning in binaries/setup.py about missing long_description_content_type
 - adds config file and manifest for release-please, including an explicit sha1 for the latest release to prevent it from going back too far in commit logs (this line will get removed after the next release)


avelad (approved):

streamer/__init__.py
```diff
@@ -1,3 +1,3 @@
-__version__ = '0.6.0'
+__version__ = '0.5.1'  # x-release-please-version
```
mariocynicys, line 1: why not start from a later version?

.github/workflows/release-please.yaml
```diff
@@ -0,0 +1,88 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+      tag_name: ${{ steps.release.outputs.tag_name }}
+      patch: ${{ steps.release.outputs.patch }}
+
+    steps:
+      # Create/update release PR
+      - uses: googleapis/release-please-action@v4
+        id: release
```
mariocynicys, line 19: does this mean we will have a release PR with every push to `main`? should it be dispatch only instead?

joeyparrish: Yes and no.  There will be a release PR after every conventional commits change that matters (fix & feat only IIRC).  We don't have to merge it right away, though.  If we merge additional changes, the release PR gets updated (version number, changelog).

The only things in the release PR are changelog and version number updates.  The version number update is based on the commits (fix only, bump patch number; feat, bump minor number; feat! (breaking change), bump major number).

We only merge the release PR when we are ready to release.  After merging the release PR, the actual release steps will execute: create GitHub release, attach artifacts to release, publish to PyPi.

streamer/__init__.py
joeyparrish: 0.5.1 was the latest we released.  The release tool (release-please) expects things to be synced (GitHub releases API, numbers in source, etc).  After I merge this, release-please will create a release PR that updates this to 1.0.0.

PR #186, opened 2024-11-05
ci: Run "Build and Test" every midnight PST / 8am UTC

avelad:


mariocynicys:
why though?

avelad:
> why though?

I think it's good to have a process that runs every day to know that nothing has broken due to something outside of this project. In Shaka Player for example this is very useful because when some browsers are updated things can break.

joeyparrish (approved):

PR #187, opened 2024-11-05
feat: Add support for FLAC codec

avelad:


streamer/bitrate_configuration.py
```diff
@@ -154,6 +155,7 @@ def _sortable_properties(self) -> Tuple[float]:
       'opus': '32k',
       'ac3': '96k',
       'eac3': '48k',
+      'flac': '705k',
```
avelad, line 158: I agree to review these values ​​if you consider it.

joeyparrish: Does FFmpeg even honor them?  FLAC is lossless, so we don't really control the bitrate, do we?

avelad: It doesn't control it, but we need to add something here to make all our logic work.

joeyparrish: Understood.  Then these numbers are fine.

joeyparrish (commented):
I'd like to merge https://github.com/shaka-project/shaka-streamer/pull/189, retest and merge https://github.com/shaka-project/shaka-streamer/pull/188, then retest and merge this.

After all of those are merged, I want to release 1.0.0.

PR #199, opened 2024-11-08
ci: Enable testing on Linux arm64

### LINKED ISSUE #180: Test on Linux arm64
Tests won't run on Linux arm64 due to a lack of Chrome builds for that platform to run the tests cases.

If may be feasible to run these cases on a different browser with a Karma launcher, but flatpak, snap, etc are not an option in our dockerized self-hosted runner environment.

joeyparrish:
Closes #180

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -120,9 +120,10 @@ jobs:
           sudo apt -y install chromium
 
           # Running inside a Docker container, we need to kill the sandbox.
+          # Heredocs interpolate variables, so escape the dollar sign below.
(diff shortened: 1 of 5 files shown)


joeyparrish:
The arm64 tests fail here because they haven't picked up the workflow changes to install Chromium.  (An effect of the "pull_request_target" trigger, meant to prevent attackers from maliciously changing a sensitive workflow.)

Here is a successful test run from my fork:

https://github.com/joeyparrish/shaka-streamer/actions/runs/11750652059/job/32739285713

avelad (approved):

PR #205, opened 2024-11-11
feat: Add generate_iframe_playlist option

avelad:


mariocynicys:
> iframe_playlist_name: The optional HLS I-Frames only playlist file to create. Usually ends with ‘.m3u8’, and is relative to hls_master_playlist_output (see below). Should only be set for video streams. If unspecified, no I-Frames only playlist is created.

Quoting: https://shaka-project.github.io/shaka-packager/html/tutorials/hls.html

Seems like this should only be set for video output only.

avelad:
> > iframe_playlist_name: The optional HLS I-Frames only playlist file to create. Usually ends with ‘.m3u8’, and is relative to hls_master_playlist_output (see below). Should only be set for video streams. If unspecified, no I-Frames only playlist is created.
> 
> Quoting: https://shaka-project.github.io/shaka-packager/html/tutorials/hls.html
> 
> Seems like this should only be set for video output only.

But I check the type of stream to make sure it is of the VIDEO type so that it only appears in the video.
imagen

streamer/packager_node.py
```diff
@@ -141,6 +141,9 @@ def _setup_stream(self, stream: OutputStream) -> str:
     if stream.type == MediaType.AUDIO:
       dict['hls_group_id'] = str(cast(AudioCodec, stream.codec).value)
 
+    if stream.type == MediaType.VIDEO and self._pipeline_config.generate_iframe_playlist:
+      dict['iframe_playlist_name'] =  stream.get_identification() + '.m3u8'
```
mariocynicys, line 145: can we also have this outputing to an `iframes/` folder or even prepend the file name with `iframe_`.

we use `hls.m3u8` for the master playlist (variable)
`dash.mpd` for dash manifest (variable)
`stream_xx.m3u8` for m3u8 playlists
`video_360p_210k_av1_2.mp4` and such for the media files. the iframe files will thus get in between those.

mariocynicys:
> But I check the type of stream to make sure it is of the VIDEO type so that it only appears in the video.

my bad, i should get glasses xD

streamer/packager_node.py
avelad: Done --> https://github.com/shaka-project/shaka-streamer/pull/205/commits/ad675b0032386baae5328d83f7add69d8d333243

mariocynicys (approved):
LGTM :)
Thanks

PR #208, opened 2024-11-11
feat: Add support for forced subtitles

avelad:


streamer/autodetect.py
```diff
@@ -173,3 +173,13 @@ def get_channel_layout(input: Input) -> Optional[AudioChannelLayoutName]:
       return bucket.get_key()
 
   return None
+
+def get_forced_subttitle(input: Input) -> bool:
+  """Returns the autodetected channel count of the input."""
```
mariocynicys, line 178: unrelated doc comment

streamer/output_stream.py
```diff
@@ -36,6 +36,7 @@ def __init__(self,
     self.input: Input = input
     self.features: Dict[str, str] = {}
     self.codec: Union[AudioCodec, VideoCodec, None] = codec
+    self.forced_subtitle = False;
```
mariocynicys, line 39: isn't that already in `Input`? why repeat it again here if we could use `self.input.forced_subtitles` when we need to query the value

avelad: I have simplified the code, and I have also followed your suggestion since this only applies to subtitles.

mariocynicys: well, i mean that `self.input` declared 3 lines above already has this data, so we don't need to copy it further in output stream. so `forced_subtitle` field here isn't really that useful.

streamer/packager_node.py
```diff
@@ -144,6 +144,9 @@ def _setup_stream(self, stream: OutputStream) -> str:
     if stream.input.drm_label:
       dict['drm_label'] = stream.input.drm_label
 
+    if stream.type == MediaType.TEXT and stream.forced_subtitle:
```
mariocynicys, line 147: and instead here we could use `stream.input.forced_subtitles` to query this value.

streamer/output_stream.py
```diff
@@ -155,6 +156,8 @@ def __init__(self,
     super().__init__(MediaType.TEXT, input, codec, pipe_dir,
                      skip_transcoding, pipe_suffix='.vtt')
 
+    self.forced_subtitle = input.forced_subtitle
+
```
mariocynicys, line 160: and remove this as well ofc.

avelad: Done!

streamer/packager_node.py
avelad: Done!

streamer/output_stream.py
avelad: Done!

mariocynicys (approved):
LGTM!
Thanks :)

Issue #216, opened 2024-12-02
Workflows disabled pending vulnerability investigation


joeyparrish:
A vulnerable workflow exposed this repo to risk of manipulation. All GitHub Actions have been disabled pending investigation of this vulnerability.

Existing releases, tags, and branches are clean and have not been poisoned. Binaries released via PyPi are also clean.

joeyparrish:
Default permissions for workflows have been fixed organization-wide.  Some workflows now need to have permissions added selectively.

joeyparrish:
Workflows have been fully audited and re-enabled for this repository.

A full discussion of the workflow vulnerability will follow soon.

joeyparrish:
# TL;DR

Above all else, I want to emphasize this:

A vulnerability was found in one of our **workflows**, not our **source code** or **binaries**.  The vulnerability was discovered by a researcher who reported it to Google and **did not damage anything**.  Workflows were suspended within hours, and no evidence of compromise has been found in any repos in the weeks I've been working on this.

If you want more details, read on.  Questions are welcome.


# The main vulnerability

The vulnerability relied on two main things:

1. The default GITHUB_TOKEN permissions were read-write
2. There was a way to inject code into the workflow in context of that token

Versions of those two things were found in several other repos across shaka-project.

As soon as the exploit PR was noticed on shaka-streamer, workflows were paused on shaka-streamer and static-ffmpeg-binaries (both used the exact same pattern).  This gave me time to investigate the PR, what it did, and how.

The exploit PR replaced `build-matrix.json` with `build-matrix.json.js`, which allowed arbitrary code to execute in context of the workflow.  This is because the nodejs code `require("./build-matrix.json")` triggers a search that first looks for the JSON (which will be automatically read and parsed), then looks for that as the name of a local module (which will be executed).

Because the default token was read-write, the PR author was able to use that token to make live calls to the GitHub API to create a new release as a proof of concept.  They could just as easily have modified an existing release, replaced attached binaries on that release, pushed a new branch, pushed to an existing unprotected branch, or created or modified tags.

The `main` branch was protected, so they would not have been able to push code directly to `main` without a merged PR.  The PR author was a security researcher who did not intend us any harm, and I have verified that no existing releases, tags, or binaries were tampered with.


# Mitigations

## Reduce default token permissions

The repositories and the organization itself all had their default token permissions set to read-write, as in this screenshot:

[...]

PR #219, opened 2024-12-16
ci: Read build matrix JSON explicitly

joeyparrish:
Because we used require() to read build-matrix.json, the file could be replaced with build-matrix.json.js, allowing code injection into our CI pipelines. This fixes this vulnerability by reading the JSON text with the fs module, then explicitly parsing it, rather than relying on require().

This exploit was discovered by a researcher, and the researcher's activity was spotted within hours.  Workflows were immediately suspended.  No evidence has been found of any tampering in this repository or its releases.

Issue #216

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -55,7 +55,8 @@ jobs:
 
           // Use ENABLE_SELF_HOSTED to decide what the build matrix below
           // should include.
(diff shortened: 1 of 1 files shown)


joeyparrish:
Workflows are still suspended, but functionality has been verified on my fork.

mariocynicys (approved):
LGTM 
thanks

PR #227, opened 2025-01-15
perf: Use a process pool to improve cloud upload performance

joeyparrish:
Python's Global Interpreter Lock (GIL) keeps multiple threads from performing as well as they could.  Although the HTTP server is threaded, those threads are not able to upload to cloud storage with the degree of parallelism you might expect.  This leads to an overall drop in the pipeline throughput, and live streams get behind.

This change refactors cloud uploads to use a pool of subprocesses.  Data from the threaded HTTP server is sent via IPC to a subprocess from a pool.  Several subprocesses then upload in parallel, while the main process can continue serving HTTP requests.

The refactor also splits up a large, unreadable module into smaller ones that are (hopefully) clearer.

The IPC mechanism provided by the multiprocessing module has a 64kB buffer, so if these queues back up, this will be reflected in the overall throughput rate logged by ffmpeg.  This is good.  If the system were willing to consume more memory to buffer waiting uploads, the pipeline might show 1x throughput right up until it dies with an out-of-memory error.

The speed-up is significant.  Issues with throughput that I previously attributed to encoding time were actually due to inefficiencies in upload.

### DIFF
--- a/streamer/controller_node.py
+++ b/streamer/controller_node.py
@@ -186,7 +186,20 @@ class ControllerNode(object):
             'For cloud uploads, the pipeline segment_per_file setting ' +
             'must be set to True!')
 
-      upload_proxy = ProxyNode.create(output_location)
+      # These are estimates.  Without constructing the pipeline nodes, we don't
+      # know precisely how many output streams there will be.  This is a good
+      # guess, though.
+      num_video_streams = (
+          len(self._pipeline_config.resolutions) *
+          len(self._pipeline_config.video_codecs))
+      num_audio_streams = (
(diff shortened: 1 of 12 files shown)


streamer/controller_node.py
```diff
@@ -315,7 +328,7 @@ def _append_nodes_for_inputs_list(self, inputs: List[Input],
                                       outputs,
                                       index,
                                       self.hermetic_ffmpeg))
-    
+
```
avelad, line 331: Thanks!

avelad (approved):

PR #229, opened 2025-01-15
docs: Fix HEVC sample config

joeyparrish:
The key recognized by shaka-streamer is "hevc", not "h265".


avelad (approved):
LGTM, I think we should add AV1 as well to serve as an example.

PR #232, opened 2025-01-16
ci: Stop using self-hosted Linux/arm64 runners

joeyparrish:
These are no longer required since GitHub launched their own Linux/arm64 runners.

See https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -122,9 +122,14 @@ jobs:
           sudo apt -y install chromium
 
           # Running inside a Docker container, we need to kill the sandbox.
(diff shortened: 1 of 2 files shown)


joeyparrish:
The arm64 failures are due to a missing environment variable I added to the workflow.  However, because we're using `pull_request_target` instead of the `pull_request` trigger, it will only run the version of the workflow from `main`.  I'll address this in a follow-up change soon.

avelad (approved):

theodab (approved):

mariocynicys (approved):

PR #272, opened 2026-05-12
fix: Fix detection of forced subtitles

avelad:
In Python, the bool() function returns True for any non-empty string. This created a bug where any value provided in the configuration (like "0", "False", or "None") would be evaluated as True simply because the string wasn't empty.

By changing it to == '1', we are implementing a strict check. Now, the subtitle track will only be marked as "forced" if the value is explicitly "1".

### DIFF
--- a/streamer/autodetect.py
+++ b/streamer/autodetect.py
@@ -182,4 +182,4 @@ def get_forced_subttitle(input: Input) -> bool:
   if forced_subttitle_string is None:
     return False
 
-  return bool(forced_subttitle_string)
+  return forced_subttitle_string == '1'
--- a/streamer/autodetect.py
+++ b/streamer/autodetect.py
@@ -174,12 +174,12 @@ def get_channel_layout(input: Input) -> Optional[AudioChannelLayoutName]:
 
   return None
 
-def get_forced_subttitle(input: Input) -> bool:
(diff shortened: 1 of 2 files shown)


joeyparrish (approved):

streamer/autodetect.py
```diff
@@ -182,4 +182,4 @@ def get_forced_subttitle(input: Input) -> bool:
   if forced_subttitle_string is None:
     return False
 
-  return bool(forced_subttitle_string)
+  return forced_subttitle_string == '1'
```
joeyparrish: I would love to see a comment above this showing an example of ffprobe's output, and explaining that it's exactly the string "1" we need to look for.

joeyparrish (approved):

streamer/autodetect.py
avelad: Done!

joeyparrish: Looks good.  I just spotted an unrelated typo (subttitle => subtitle) in this function, but it doesn't have to be fixed in this PR.

joeyparrish (approved):

streamer/autodetect.py
avelad: Done!

PR #274, opened 2026-05-12
feat: Improve HLS track naming and fix forced subtitles collision

avelad:


avelad:
@joeyparrish can you review it? Thanks!

joeyparrish (approved):

PR #279, opened 2026-07-09
test: Bind the end-to-end test server to a free port

### LINKED ISSUE #261: Flaky test runs
Ex: https://github.com/shaka-project/shaka-streamer/actions/runs/23347753424/job/67917715414?pr=260

```
Chrome 145.0.0.0 (Mac OS 10.15.7): Executed 57 of 58 (1 FAILED) (skipped 1) (2 mins 37.664 secs / 2 mins 37.626 secs)
TOTAL: 1 FAILED, 56 SUCCESS


1) fails when extra fields are present
     Shaka Streamer
     Expected a promise to be rejected with
         <jasmine.objectContaining(Object({ error_type: 'UnrecognizedField', field_name: 'foo' }))>
         but it was rejected with TypeError: Failed to fetch.
Error
    at <Jasmine>
    at UserContext.<anonymous> (tests.js:247:10)
    at <Jasmine>
```

joeyparrish:
The first end-to-end test ('fails when extra fields are present') failed intermittently on macOS with 'TypeError: Failed to fetch', without ever reaching the config validation it was meant to exercise, so the flake was in the test harness rather than in Shaka Streamer itself.

What we know:
- The failure was macOS-only and always hit the first request to the Flask test server, which used Flask's default port 5000.
- On macOS, the ControlCenter/AirPlay Receiver service also listens on port 5000.
- Binding the server to an OS-assigned free port instead of 5000 stops the flake: the fix ran green across several repeated macOS CI runs.

What we do NOT know:
- The exact interaction on the CI runners.  Our server and AirPlay can both bind port 5000 via SO_REUSEADDR, and locally a 127.0.0.1:5000 request still reaches our server, so we could not reproduce the precise failure mode off the runners.  An earlier attempt to force IPv4 (using 127.0.0.1 instead of 'localhost') did not help, which is what pointed us at the port rather than address resolution.

So this is an empirically-verified fix without a fully pinned-down
mechanism.  Using a free port also avoids any future collision with an unrelated service.

The server now binds via werkzeug.serving.make_server so we can read the chosen port back before serving (no window for another process to steal it), and that port is passed to the browser tests through the Karma config.

Fixes #261

### DIFF
--- a/run_end_to_end_tests.py
+++ b/run_end_to_end_tests.py
@@ -19,6 +19,7 @@ browser."""
 
 import argparse
 import flask
+from werkzeug.serving import make_server
 import glob
 import json
 import logging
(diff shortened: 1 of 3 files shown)


joeyparrish:
Looks like this didn't fix it.  Converting to a draft while I iterate on the details in CI.

joeyparrish:
Rewritten (code & PR title & description) and verified to fix the most common test flake.  Ready for review.

avelad (approved):

